PR #8, opened 2021-06-03
Add amplitude session plugin sample

migs647:
Test Plan:

- Verify sessionID starts as -1
    - Verify when the application starts, sessionID is -1 by placing a breakpoint in startTimer()
    
- Verify timer fires
    - Verify timer fires after 5 min (+/- 1 second)
    - Verify sessionID changes after fire
    - Verify amplitude integration payload session_id is updated

- Verify sessionID once set is 13 digits
    - Set a breakpoint at line 81 (in startTimer()) to verify the sessionID has changed and is 13 digits. 

Screen Shot 2021-06-03 at 2 15 33 PM
Screen Shot 2021-06-03 at 2 15 44 PM
Screen Shot 2021-06-03 at 2 16 13 PM
Screen Shot 2021-06-03 at 2 17 43 PM
Screen Shot 2021-06-03 at 2 19 04 PM
Screen Shot 2021-06-03 at 2 19 19 PM


samples/kotlin-android-app-destinations/src/main/java/com/segment/analytics/destinations/plugins/AmplitudeSession.kt
```diff
@@ -0,0 +1,95 @@
...
+import com.segment.analytics.platform.DestinationPlugin
+import com.segment.analytics.platform.plugins.LogType
+import com.segment.analytics.platform.plugins.log
+import com.segment.analytics.utilities.putIntegrations
+import java.time.Instant
+import java.util.*
+import kotlin.concurrent.schedule
+
+// A Destination plugin that adds session tracking to Amplitude cloud mode.
+class AmplitudeSession() : DestinationPlugin(), LifecycleObserver {
```
prayansh, line 18: i think we should use the `AndroidLifecycle` interface instead of `LifecycleObserver`

samples/kotlin-android-app-destinations/src/main/java/com/segment/analytics/destinations/plugins/AmplitudeSession.kt
```diff
@@ -0,0 +1,95 @@
...
+        return payload
+    }
+
+    override fun alias(payload: AliasEvent): BaseEvent? {
+        insertSession(payload)
+        return payload
+    }
+
+    @OnLifecycleEvent(Lifecycle.Event.ON_STOP)
+    fun onBackground() {
```
prayansh, line 67: this can be replaced with the `onActivityStopped()` callback from `AndroidLifecycle`

samples/kotlin-android-app-destinations/src/main/java/com/segment/analytics/destinations/plugins/AmplitudeSession.kt
```diff
@@ -0,0 +1,95 @@
...
+        log(LogType.INFO, "background", null)
+        stopTimer()
+    }
+
+    @OnLifecycleEvent(Lifecycle.Event.ON_START)
+    fun onForeground() {
```
prayansh, line 72: this can be replaced with the `onActivityStarted()`/`onActivityResumed()` callback from `AndroidLifecycle`

samples/kotlin-android-app-destinations/src/main/java/com/segment/analytics/destinations/plugins/WebhookPlugin.kt
```diff
@@ -1,4 +1,4 @@
-package com.segment.analytics.next.plugins
+package com.segment.analytics.destinations.plugins
```
prayansh, line 1: 🙌

prayansh (changes requested):
i am not sure if using Lifecycle observer directly is the best way since we do provide hooks into the lifecycle via `AndroidLifecycle` interface for the plugins. I think we should switch to those rather than directly interacting with the Lifecycle observer

prayansh (approved):
LGTM

PR #10, opened 2021-06-15
Startup Queue Logic

prayansh:
## Purpose
If the analytics instance has not yet received a settings update, then most destinations arent ready for the incoming events. This change will allow us to pause the analytics instance and **queue up** any incoming events until the instance has received settings.

## Approach
We add a new plugin called `StartupQueue` which queues up events (in memory) until the analytics instance is `running`. The `running` state is managed via `sovran-kotlin` and is initialized as false when first started and only set to true when we have successfully fetched the settings. The running state is also set to false in between settings updates and then re-set to true once the update is finished

### DIFF
--- a/analytics-kotlin/src/main/java/com/segment/analytics/Settings.kt
+++ b/analytics-kotlin/src/main/java/com/segment/analytics/Settings.kt
@@ -8,7 +8,6 @@ import kotlinx.serialization.decodeFromString
 import kotlinx.serialization.json.Json
 import kotlinx.serialization.json.JsonObject
 import java.io.BufferedReader
-import java.util.zip.GZIPInputStream
 
 @Serializable
 data class Settings(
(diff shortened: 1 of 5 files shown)


analytics-kotlin/src/main/java/com/segment/analytics/platform/plugins/StartupQueue.kt
```diff
@@ -0,0 +1,63 @@
...
+            subscriber = this,
+            stateClazz = System::class,
+            initialState = true,
+            handler = this::runningUpdate
+        )
+    }
+
+    override fun execute(event: BaseEvent): BaseEvent? {
+        if (!started.get()) {
+            analytics.log("PRAY-message-queuing")
```
bsneed, line 33: remove the log's, or take your name out of them. :D

bsneed (approved):

analytics-kotlin/src/main/java/com/segment/analytics/platform/plugins/StartupQueue.kt
prayansh: whooops 😅

analytics-kotlin/src/main/java/com/segment/analytics/Settings.kt
```diff
@@ -17,25 +16,42 @@ data class Settings(
     var edgeFunction: JsonObject = emptyJsonObject,
 )
 
+internal fun Analytics.update(settings: Settings) {
+    timeline.applyClosure { plugin ->
+        // tell all top level plugins to update.
+        // For destination plugins they auto-handle propagation to sub-plugins
+        plugin.update(settings)
```
migs647, line 23: Do we care to filter destination plugins or can all plugins have "started" state?

migs647 (approved):
LGTM

analytics-kotlin/src/main/java/com/segment/analytics/Settings.kt
prayansh: i think all should have it, filtering the `started` state to only destinations may make things difficult to understand in the long term

PR #12, opened 2021-06-16
Referrer Property in Deep Link Opened

prayansh:
## Purpose
Deep Link Opened events can now also collect information regarding the referrer

## Approach
Following this tutorial https://clmirror.storage.googleapis.com/codelabs/deeplink-referrer/index.html?index=..%2F..index#5, we now have the ability to track the referrer for the `Deep Link Opened` event. This value can be either the website url or the app package name that contained the deep link.

### DIFF
--- a/analytics-kotlin/src/test/java/com/segment/analytics/main/AndroidLifecyclePluginTests.kt
+++ b/analytics-kotlin/src/test/java/com/segment/analytics/main/AndroidLifecyclePluginTests.kt
@@ -305,6 +305,78 @@ class AndroidLifecyclePluginTests {
         }
     }
 
+    @Config(sdk = [22])
+    @Test
+    fun `track deep link when enabled sdk=22`() {
+        analytics.configuration.trackApplicationLifecycleEvents = false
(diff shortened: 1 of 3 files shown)


analytics-kotlin/src/test/java/com/segment/analytics/main/AndroidLifecyclePluginTests.kt
```diff
@@ -305,6 +305,80 @@ class AndroidLifecyclePluginTests {
...
+            assertEquals(buildJsonObject {
+                put("url", "app://track.com/open?utm_id=12345&gclid=abcd&nope=")
+                put("utm_id", "12345")
+                put("gclid", "abcd")
+                put("referrer", "android-app:/com.package.app")
+            }, properties)
+        }
+    }
+/*
+    Due to some complications b/w mockk and robolectric this test seems to be failing on the CI
```
prayansh, line 344: for context: this test is working on my local machine, but for some reason fails on the CI, which I believe is stemming from a problem between mockk and robolectric

migs647 (approved):
LGTM

PR #14, opened 2021-06-28
Added Firebase destination

migs647:
To test... 

Replace google-services.json file with your own to see the events on Firebase show.

### DIFF
--- a/analytics-kotlin/src/main/java/com/segment/analytics/utilities/JSON.kt
+++ b/analytics-kotlin/src/main/java/com/segment/analytics/utilities/JSON.kt
@@ -74,6 +74,7 @@ fun JsonPrimitive.toContent(): Any? {
     this.doubleOrNull?.let {
         return it
     }
+    
     return contentOrNull
 }
 
@@ -93,6 +94,19 @@ fun JsonObject.getInt(key: String): Int? = this[key]?.jsonPrimitive?.intOrNull
 fun JsonObject.getStringSet(key: String): Set<String>? =
     this[key]?.jsonArray?.map { it.jsonPrimitive.content }?.toSet()
 
-fun JsonObject.getMapSet(key: String): Set<Map<String, JsonElement>>? {
-    return this[key]?.jsonObject?.map { mapOf(it.key to it.value) }?.toSet()
+// Utility function to retrieve a map set from a jsonObject
+fun JsonObject.getMapSet(key: String): Set<Map<String, Any>>? {
+
+    val returnList: MutableList<Map<String, Any>> = mutableListOf()
+
+    this[key]?.jsonObject?.let { jsonMap ->
+        for ((mapKey, value) in jsonMap) {
+            returnList.add(mapOf(mapKey to value))
+        }
+    }
+
+    return if (returnList.isNotEmpty())
+        returnList.toSet()
+    else
+        null
 }
--- a/analytics-kotlin/src/main/java/com/segment/analytics/utilities/JSON.kt
+++ b/analytics-kotlin/src/main/java/com/segment/analytics/utilities/JSON.kt
@@ -11,7 +11,7 @@ fun JsonObjectBuilder.putAll(obj: JsonObject) {
     }
 }
 
-// Utility function to put "undefined" value in-stead of `null` when building JsonObject
+// Utility function to put "undefined" value instead of `null` when building JsonObject
 fun JsonObjectBuilder.putUndefinedIfNull(key: String, value: CharSequence?): JsonElement? =
     if (value.isNullOrEmpty()) {
         put(key, "undefined")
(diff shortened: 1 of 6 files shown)


analytics-kotlin/src/main/java/com/segment/analytics/utilities/JSON.kt
```diff
@@ -78,11 +78,21 @@ fun JsonPrimitive.toContent(): Any? {
...
+fun JsonObject.getDouble(key: String): Double? = this[key]?.jsonPrimitive?.doubleOrNull
+
+// Utility function to retrieve a int value from a jsonObject
+fun JsonObject.getInt(key: String): Int? = this[key]?.jsonPrimitive?.intOrNull
+
 // Utility function to retrieve a string set (from jsonArray) from a jsonObject
 fun JsonObject.getStringSet(key: String): Set<String>? =
     this[key]?.jsonArray?.map { it.jsonPrimitive.content }?.toSet()
+
+fun JsonObject.getMapSet(key: String): Set<Map<String, JsonElement>>? {
```
prayansh, line 96: any reason why we are returning a `Set<Map<String, JsonElement>>` instead of `Set<Map<String, Any>>`?

prayansh: could potentially do this 
```
fun JsonObject.getMapSet(key: String): Set<Map<String, Any?>>? =
    this[key]?.jsonObject?.map { mapOf(it.key to it.value.jsonPrimitive.toContent()) }?.toSet()
```

samples/kotlin-android-app-destinations/src/main/java/com/segment/analytics/destinations/plugins/FirebaseDestination.kt
```diff
@@ -0,0 +1,236 @@
...
+        if (EVENT_MAPPER.containsKey(eventName)) {
+            eventName = EVENT_MAPPER[eventName].toString()
+        } else {
+            eventName = makeKey(eventName)
+        }
+
+        payload.properties.let {
+            val formattedProperties = formatProperties(it)
+            firebaseAnalytics.logEvent(eventName, formattedProperties)
+            analytics.log("firebaseAnalytics.logEvent($eventName, $formattedProperties")
```
prayansh, line 105: missing a trailing `)`

samples/kotlin-android-app-destinations/src/main/java/com/segment/analytics/destinations/plugins/FirebaseDestination.kt
```diff
@@ -0,0 +1,236 @@
...
+            val packageManager = activity?.packageManager ?: return
+
+            packageManager.getActivityInfo(activity.componentName, PackageManager.GET_META_DATA).let {
+                it.loadLabel(packageManager).toString().let {
+                    firebaseAnalytics.setCurrentScreen(activity, it, null)
+                    analytics.log("firebaseAnalytics.setCurrentScreen(activity, $it, null")
+                }
+            }
+        } catch (exception: PackageManager.NameNotFoundException) {
+            throw AssertionError("Activity Not Found: " + exception.toString())
```
prayansh, line 137: not sure if we want to be throwing an error or simply logging it

samples/kotlin-android-app-destinations/src/main/java/com/segment/analytics/destinations/plugins/FirebaseDestination.kt
```diff
@@ -0,0 +1,236 @@
...
+
+    override fun onActivityStopped(activity: Activity?) {
+        super.onActivityStopped(activity)
+        this.activity = null
+    }
+
+    // Private Helper Methods
+
+    // Format properties into a format needed by firebase
+    private fun formatProperties(properties: Properties): Bundle? {
```
prayansh, line 154: Ref: https://developer.android.com/kotlin/ktx/extensions-list#androidxcoreos
`androidx.core:core-ktx` has a bunch of nicer APIs specifically for kotlin, to build Bundles. Could potentially look into using those to kotlin-ify the code a bit more 😄

analytics-kotlin/src/main/java/com/segment/analytics/utilities/JSON.kt
migs647: I'm game for that

samples/kotlin-android-app-destinations/src/main/java/com/segment/analytics/destinations/plugins/FirebaseDestination.kt
migs647: Hah I was confused why it was compiling, but realized it was just in the string. Added.

samples/kotlin-android-app-destinations/src/main/java/com/segment/analytics/destinations/plugins/FirebaseDestination.kt
migs647: We were throwing on the previous destination, do we want to change behavior? Down to log it.

samples/kotlin-android-app-destinations/src/main/java/com/segment/analytics/destinations/plugins/FirebaseDestination.kt
prayansh: whoops, should have been clearer 😅

prayansh (approved):

PR #17, opened 2021-07-23
Updated to latest firebase analytics for destination

migs647:
Updated to the newest version which had some deprecations around screen tracking. Also fixed an issue with track method skipping if properties weren't included.

### DIFF
--- a/samples/kotlin-android-app-destinations/src/main/java/com/segment/analytics/destinations/plugins/FirebaseDestination.kt
+++ b/samples/kotlin-android-app-destinations/src/main/java/com/segment/analytics/destinations/plugins/FirebaseDestination.kt
@@ -103,11 +103,13 @@ class FirebaseDestination(
             eventName = makeKey(eventName)
         }
 
+        var bundledProperties: Bundle?
         payload.properties.let {
-            val formattedProperties = formatProperties(it)
-            firebaseAnalytics.logEvent(eventName, formattedProperties)
-            analytics.log("firebaseAnalytics.logEvent($eventName, $formattedProperties)")
+            bundledProperties = formatProperties(it)
         }
+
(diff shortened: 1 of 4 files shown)


samples/kotlin-android-app-destinations/src/main/java/com/segment/analytics/destinations/plugins/FirebaseDestination.kt
```diff
@@ -103,12 +103,14 @@ class FirebaseDestination(
             eventName = makeKey(eventName)
         }
 
+        var bundledProperties: Bundle?
         payload.properties.let {
-            val formattedProperties = formatProperties(it)
-            firebaseAnalytics.logEvent(eventName, formattedProperties)
-            analytics.log("firebaseAnalytics.logEvent($eventName, $formattedProperties)")
+            bundledProperties = formatProperties(it)
         }
```
prayansh, line 109: more idiomatic to write
```
val bundledProperties = payload.properties?.let {
  formatProperties(it)
}
```

prayansh (changes requested):
Lets also add the logic to replace the forbidden characters with `_` instead of filtering them out here. https://github.com/segmentio/analytics-kotlin/pull/17/files#diff-eb2eec385880c84c9b2512d6b31654185e07bd923c4544014e1bb0a94d37f7d8R223-R227

samples/kotlin-android-app-destinations/src/main/java/com/segment/analytics/destinations/plugins/FirebaseDestination.kt
migs647: Cool but it'd be ...
`bundledProperties = payload.properties?.let {
  formatProperties(it)
}`

prayansh (approved):

PR #18, opened 2021-08-03
add new helpers for destinations work

prayansh:
I am adding a bunch of new functions/apis that will aid in building new destinations.

```
+ Settings.isDestinationEnabled(name: String)
+ Settings.destinationSettings(name: String) // Will produce a typed config

+ JsonObject.transformKeys(transform: (String) -> String)
+ JsonObject.transformValuestransform: (String) -> String)
```


migs647 (approved):
Looks great

PR #19, opened 2021-08-04
Refactor plugin interface and update apis

prayansh:
Multiple changes in this PR
- Refactor plugin to not have any `name` 
- Refactor destination plugins to have `key` instead of `name`
- update the `find / remove` plugin apis
- Adding a way to disable destination plugins based on the incoming settings
- Update webhook destination to be an after plugin
- Update AmplitudeSession plugin example to be an enrichment plugin


core/src/test/kotlin/com/segment/analytics/kotlin/core/PluginTests.kt
```diff
@@ -36,7 +36,7 @@ class PluginTests {
     }
 
     @Test
-    fun `execute runs the destination timeline`() {
+    fun `plugin processes the event correctly`() {
```
prayansh, line 39: nothing to do with this change, i just realized that the name was incorrectly written

migs647 (approved):
LGTM. We may want to add the record screen views plugin and removing of the configuration flag as a delta between 1.0 and 2.0.

bsneed (approved):

PR #23, opened 2021-08-13
Add a base64 encoding implementation

### LINKED ISSUE #22: Use of unsupported Base64 API in API < 26
The Java Base64 encoder isn't supported in Android API/SDK versions < 26, but the minSDK of the project is version 16

https://github.com/segmentio/analytics-kotlin/blob/494e6e537bafbd982aef7b023b80e9c25152e4ec/core/src/main/java/com/segment/analytics/kotlin/core/HTTPClient.kt#L37

See https://developer.android.com/reference/java/util/Base64

Causes crashes like this:
```
Fatal Exception: java.lang.NoClassDefFoundError: Failed resolution of: Ljava/util/Base64;
       at com.segment.analytics.kotlin.core.HTTPClient.authorizationHeader(HTTPClient.java)
       at com.segment.analytics.kotlin.core.HTTPClientKt.createPostConnection(HTTPClientKt.java)
       at com.segment.analytics.kotlin.core.SegmentDestination.performFlush(SegmentDestination.java:3)
       at com.segment.analytics.kotlin.core.SegmentDestination$flush$$inlined$run$lambda$1.invokeSuspend(SegmentDestination.java:3)
       at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(BaseContinuationImpl.java)
       at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.java)
       at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.java)
       at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.java:5)
```

prayansh:
Closes #22 

Context: android pre API 26 does not implement `java.util.Base64` and recommends using `android.util.Base64` instead. I couldn't find a way to switch between the 2 implementations while supporting both android and JVM, so I wrote my own implementation of Base64 encoding. 

Additional changes: 
- change upload endpoint to use `/batch` instead of `/import`
- make httpclient stateful
- cache authHeader in httpClient to prevent base64 recomputation
- improve httpclient tests

### DIFF
--- a/core/src/main/java/com/segment/analytics/kotlin/core/HTTPClient.kt
+++ b/core/src/main/java/com/segment/analytics/kotlin/core/HTTPClient.kt
@@ -1,5 +1,6 @@
 package com.segment.analytics.kotlin.core
 
+import com.segment.analytics.kotlin.core.utilities.encodeToBase64
 import java.io.BufferedReader
 import java.io.Closeable
 import java.io.IOException
@@ -34,7 +35,7 @@ class HTTPClient {
 
     private fun authorizationHeader(writeKey: String): String {
         val auth = "$writeKey:"
-        return "Basic " + Base64.getEncoder().encodeToString(auth.toByteArray())
+        return "Basic ${encodeToBase64(auth)}"
     }
 
     /**
--- a/core/src/main/java/com/segment/analytics/kotlin/core/HTTPClient.kt
+++ b/core/src/main/java/com/segment/analytics/kotlin/core/HTTPClient.kt
@@ -1,5 +1,6 @@
 package com.segment.analytics.kotlin.core
 
+import com.segment.analytics.kotlin.core.Constants.LIBRARY_VERSION
 import com.segment.analytics.kotlin.core.utilities.encodeToBase64
 import java.io.BufferedReader
 import java.io.Closeable
@@ -11,8 +12,10 @@ import java.net.MalformedURLException
 import java.net.URL
 import java.util.zip.GZIPOutputStream
 
-class HTTPClient {
-    fun settings(writeKey: String): Connection {
+class HTTPClient(private val writeKey: String) {
+    internal val authHeader = authorizationHeader(writeKey)
+
+    fun settings(): Connection {
         val connection: HttpURLConnection =
             openConnection("https://cdn-settings.segment.com/v1/projects/$writeKey/settings")
         val responseCode = connection.responseCode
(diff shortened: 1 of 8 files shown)


wenxi-zeng (approved):
LGTM

core/src/main/java/com/segment/analytics/kotlin/core/HTTPClient.kt
```diff
@@ -34,7 +34,7 @@ class HTTPClient {
 
     private fun authorizationHeader(writeKey: String): String {
         val auth = "$writeKey:"
-        return "Basic " + Base64.getEncoder().encodeToString(auth.toByteArray())
+        return "Basic ${encodeToBase64(auth)}"
```
migs647, line 37: I feel like it may be a good idea to only re-encode when the writeKey changes instead of every time we ask for the authorizationHeader. Mainly because our `encodeToBase64` method is O(n^2). Thoughts?

migs647 (commented):
Just need clarity to the efficiency comment.

core/src/main/java/com/segment/analytics/kotlin/core/HTTPClient.kt
prayansh: ooh good catch, lemme see if I can add that optimization in our codebase

wenxi-zeng: correct me if I'm wrong. I think our `encodeToBase64` method is still O(n). the inner for loops are capped by 3, so worst case O(3n). but still changing it to only re-encode when writeKey changes is a good idea! XD

prayansh: yea i havent done the full analysis, but wenxi might be right, its of the order O(n) rather than O(n^2). But anyways caching the header is a good idea since it does not actually change over time.

migs647: I see it as O(n^2) because We have a while loop and inside the while loop 3n (3 for loops).

migs647 (approved):
LGTM

core/src/main/java/com/segment/analytics/kotlin/core/HTTPClient.kt
wenxi-zeng: ah. gotcha. yeah. that also makes sense. I was thinking the while loop takes m, the 3 inner loops each takes n (where n <= 3), thus the complexity is O(m*3n) <= O(m*3*3) = O(9m) = O(m). but it also make sense to treat the inner loop n as a variable, and indeed it is. so O(m*3n) is more accurate. 😄

PR #24, opened 2021-08-19
Add new mapTransform function

prayansh:
In our efforts to provide convenience APIs, we have added a new function `JsonObject.mapTransform(Map<String, String>,((String, JsonElement) -> JsonElement)?)` that allows you to easily map keys and transform values based on a closure. Example usage:
```kotlin
 properties.mapTransform(mapOf("x" to "y")) { newKey, value ->
    var newVal = value
    if (newKey == "y") {
        // ... do something
    }
    newVal
}
```

### DIFF
--- a/core/src/main/java/com/segment/analytics/kotlin/core/utilities/JSON.kt
+++ b/core/src/main/java/com/segment/analytics/kotlin/core/utilities/JSON.kt
@@ -1,6 +1,5 @@
 package com.segment.analytics.kotlin.core.utilities
 
-import kotlinx.serialization.json.Json
 import kotlinx.serialization.json.JsonArray
 import kotlinx.serialization.json.JsonElement
 import kotlinx.serialization.json.JsonObject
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/segment/analytics/kotlin/core/utilities/JSON.kt
```diff
@@ -83,6 +86,49 @@ fun JsonObject.getMapSet(key: String): Set<Map<String, Any>>? {
...
+        // does this key1 have a mapping?
+        keyMapper[key]?.let { mappedKey ->
+            newKey = mappedKey
+        }
+
+        // is this value a dictionary?
+        if (value is JsonObject) {
+            // if so, lets recurse...
+            newVal = value.mapTransform(keyMapper, valueTransform)
+        } else if (value is JsonArray) {
```
wenxi-zeng, line 107: it's probably better to have an extension method `mapTransform` for `JsonArray` and wrap this block of `else if` into that extension method for readability. otherwise `newVal` and `newValue` is a bit confusing.

wenxi-zeng: rest LGTM :)

prayansh: ooh good catch. will do

wenxi-zeng (approved):

PR #29, opened 2021-09-01
Json Utility improvements & Re-add Alias Event

prayansh:
- Adding a few more utility functions to make working with Json types simpler
- Add new pre-made `Json` constructors to avoid recreation of special `Json` formatters on each use
- Added tests for new functions
- Improved code quality of tests
- Re-added Alias event APIs
- Add new safe `JsonElement -> (JsonObject | JsonArray | JsonPrimitive)` conversions
- Update usages of json converters to safe ones

### DIFF
--- a/core/src/test/kotlin/com/segment/analytics/kotlin/core/JSONTests.kt
+++ b/core/src/test/kotlin/com/segment/analytics/kotlin/core/JSONTests.kt
@@ -4,7 +4,7 @@ import com.segment.analytics.kotlin.core.utilities.getBoolean
 import com.segment.analytics.kotlin.core.utilities.getDouble
 import com.segment.analytics.kotlin.core.utilities.getInt
 import com.segment.analytics.kotlin.core.utilities.getLong
-import com.segment.analytics.kotlin.core.utilities.getMapSet
+import com.segment.analytics.kotlin.core.utilities.getMapList
 import com.segment.analytics.kotlin.core.utilities.getString
 import com.segment.analytics.kotlin.core.utilities.getStringSet
 import com.segment.analytics.kotlin.core.utilities.mapTransform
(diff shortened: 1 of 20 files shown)


core/src/test/kotlin/com/segment/analytics/kotlin/core/JSONTests.kt
```diff
@@ -3,503 +3,712 @@ package com.segment.analytics.kotlin.core
...
+            val jsonObject = JsonObject(mapOf("keyed" to JsonPrimitive("18")))
 
-        try {
-            val keyedValue = jsonObject.getInt("keyed")
-            assertTrue(keyedValue?.equals(18) ?: false)
-        } catch (e: Exception) {
-            fail("Should not throw when not int primitive")
+            try {
+                val keyedValue = jsonObject.getInt("keyed")
+                assertTrue(keyedValue?.equals(18) ?: false)
```
wenxi-zeng, line 157: nit: it's probably better to use `assertEquals`. same applies to other similar usages in this file

samples/kotlin-android-app-destinations/src/main/java/com/segment/analytics/destinations/EventFragment.kt
```diff
@@ -109,10 +109,12 @@ class EventFragment(val type: EventType, val analytics: Analytics) : Fragment()
         container.addView(view)
     }
 
-    private inline fun <reified T : BaseEvent> eventStr(event: T) = Json {
+    private val Json = Json {
```
wenxi-zeng, line 112: nit: what's the consideration behind declaring a separate variable? it seems to be only used once and it's private. I think it's probably better to make it a functional level variable:
```kotlin
    private inline fun <reified T : BaseEvent> eventStr(event: T): String{
        val json = Json {
            prettyPrint = true
            encodeDefaults = true
        }
        
        return json.encodeToString(event)
    }
```

core/src/test/kotlin/com/segment/analytics/kotlin/core/JSONTests.kt
prayansh: yup good catch, i copied the language from other parts. will update everywhere :)

samples/kotlin-android-app-destinations/src/main/java/com/segment/analytics/destinations/EventFragment.kt
prayansh: this would put us in the same place as before. What we are trying to do is to avoid constructing `Json { ... }` each time, and having it inside the function will in turn mean that everytime `evenStr` is called, we create a new copy of that object.

wenxi-zeng: ah. totally make sense 👍

wenxi-zeng (approved):
LGTM

core/src/test/kotlin/com/segment/analytics/kotlin/core/JSONTests.kt
```diff
@@ -3,503 +3,734 @@ package com.segment.analytics.kotlin.core
...
 import org.junit.jupiter.api.Assertions.fail
+import org.junit.jupiter.api.Nested
 import org.junit.jupiter.api.Test
 import org.junit.jupiter.api.TestInstance
-import java.lang.IllegalArgumentException
+import org.junit.jupiter.params.ParameterizedTest
+import org.junit.jupiter.params.provider.ValueSource
 
 @TestInstance(TestInstance.Lifecycle.PER_CLASS)
 class JSONTests {
```
prayansh, line 39: @migs647, I made changes here to make it more kotlin-like, if u wanna take a look

migs647 (approved):
LGTM

PR #30, opened 2021-09-02
LIBMOBILE-529 java compatibility

wenxi-zeng:
**Summary**
* add `JavaAnalytics` that implements adapter pattern to provide java compatibility accessing some of the `Analytics` methods
* add `ConfigurationBuilder` that implements builder pattern to help build `Configuration` object in java
* add `Builder` which contains a seres of convenient methods that helps building `Kotlin JsonObject` in java
* add `JsonSerializable` that helps converting user's own classes to be serializable and make the use of `Analytics` api simple
* add more convenient functions in `EventManipulationFunctions.kt` to provide java compatibility. rename it to be `EventTransformer` in java 
* add `JAVA_COMPAT.md` to shows the usage of `Analytics` in pure java codebase
* replace outdated `README.md` info to current
* add unit tests for the new compatibility classes
* add `alias` and corresponding readme contents


README.md
```diff
@@ -7,26 +7,32 @@ NOTE: This project is currently in the Pilot phase and is covered by Segment's [
...
+    - [remove](#remove)
+    - [flush](#flush)
+  - [Plugin Architecture](#plugin-architecture)
     - [Fundamentals](#fundamentals)
-    - [Advanced Concepts](#advanced-concepts)
-- [Contributing](#contributing)
-- [Code of Conduct](#code-of-conduct)
-- [License](#license)
+    - [Advanced concepts](#advanced-concepts)
+  - [Contributing](#contributing)
```
prayansh, line 33: whats the reason for this change?

core/src/main/java/com/segment/analytics/kotlin/core/utilities/EventManipulationFunctions.kt
```diff
@@ -18,59 +18,75 @@ fun BaseEvent.disableIntegration(integrationName: String): BaseEvent {
...
+fun BaseEvent.putInContext(key: String, jsonElement: JsonElement): BaseEvent {
     context = buildJsonObject {
         putAll(context)
-        put(key, Json.encodeToJsonElement(serializationStrategy, value))
+        put(key, jsonElement)
     }
     return this
 }
 
+fun BaseEvent.putInContext(key: String, value: String?) = putInContext(key, JsonPrimitive(value))
```
prayansh, line 48: ❤️

core/src/test/kotlin/com/segment/analytics/kotlin/core/compat/BuildersTest.kt
```diff
@@ -0,0 +1,224 @@
+package com.segment.analytics.kotlin.core.compat
+
+import kotlinx.serialization.json.*
+import org.junit.jupiter.api.Assertions.assertEquals
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.TestInstance
+
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+internal class BuildersTest {
```
prayansh, line 9: do u think its better to write these tests in java? or is it the same to have it in kotlin?

core/src/test/kotlin/com/segment/analytics/kotlin/core/compat/JavaAnalyticsTest.kt
```diff
@@ -0,0 +1,399 @@
...
+import kotlinx.serialization.json.JsonObject
+import kotlinx.serialization.json.buildJsonObject
+import kotlinx.serialization.json.put
+import org.junit.jupiter.api.Assertions.*
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Nested
+import org.junit.jupiter.api.Test
+import java.util.function.Consumer
+
+internal class JavaAnalyticsTest {
```
prayansh, line 23: same thought here. would it be better to have this in java?

prayansh (approved):
LGTM. couple of considerations to ponder upon

README.md
wenxi-zeng: I used a vcs plugin to update the table of content. it changes the content automatically. it's mainly for adding `alias` to table of content

core/src/test/kotlin/com/segment/analytics/kotlin/core/compat/BuildersTest.kt
wenxi-zeng: yeah. that's a good callout. I actually thought about it. the test cases here are mainly for unit tests. so the function invocations are the same in both java and kotlin. I use the way we do it in kotlin as the `expected` result, and the way we do it in java as the `actual` result, to verify they are equivalent. I'm thinking adding integration test cases in `jvm-example` to verify the actual usage in java, so we keep all the code in `core` purely kotlin.

PR #32, opened 2021-09-02
minimize the number of times of open/close file

wenxi-zeng:
* minimize the number of times of open/close file
* add shutdown hook to close file when app shuts down

### DIFF
--- a/core/src/main/java/com/segment/analytics/kotlin/core/utilities/EventsFileManager.kt
+++ b/core/src/main/java/com/segment/analytics/kotlin/core/utilities/EventsFileManager.kt
@@ -1,5 +1,6 @@
 package com.segment.analytics.kotlin.core.utilities
 
+import kotlinx.coroutines.runBlocking
 import java.io.File
 import java.io.FileOutputStream
 import java.time.Instant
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/segment/analytics/kotlin/core/utilities/EventsFileManager.kt
```diff
@@ -120,6 +124,8 @@ class EventsFileManager(
         val contents = """],"sentAt":"${Instant.now()}"}"""
```
wenxi-zeng, line 124: @prayansh wonder if we should replace `Instant.now()` with `SimpleDateFormat` to be back compatible. this call is only available on android api 26 or above

migs647 (approved):
This is a step in the right direction. LGTM for now

core/src/main/java/com/segment/analytics/kotlin/core/utilities/EventsFileManager.kt
prayansh: we dont need to do that because we are using the `coreLibraryDesguaring` apis, which allows us to use Instant in older API levels.

PR #34, opened 2021-09-08
SegmentDestination refactor and add `cdnHost` config option

prayansh:
Summary of changes
- General code cleanup
- Add new config option `cdnHost` enabling users to modify from where settings are fetched
- Refactor Segment destination to new template format


wenxi-zeng (approved):
LGTM. can you also add a builder function to `ConfigurationBuilder` for `cdnHost`?

prayansh:
good call. will add that

wenxi-zeng (approved):

PR #36, opened 2021-09-13
keep dispatcher and scope config internal

wenxi-zeng:
**Changes:**
* remove dispatchers and scope from `Configuration` constructors, keep them internal
* fix broken tests

**Additional Changes**
* temporary fix the blocking issue caused by sovran

### DIFF
--- a/android/src/test/java/com/segment/analytics/kotlin/android/AndroidLifecyclePluginTests.kt
+++ b/android/src/test/java/com/segment/analytics/kotlin/android/AndroidLifecyclePluginTests.kt
@@ -13,8 +13,6 @@ import com.segment.analytics.kotlin.core.platform.Plugin
 import com.segment.analytics.kotlin.android.plugins.AndroidLifecycle
 import com.segment.analytics.kotlin.android.plugins.AndroidLifecyclePlugin
 import io.mockk.*
-import kotlinx.coroutines.test.TestCoroutineDispatcher
-import kotlinx.coroutines.test.TestCoroutineScope
 import kotlinx.serialization.json.buildJsonObject
 import kotlinx.serialization.json.put
 import org.junit.Assert.*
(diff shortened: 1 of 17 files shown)


android/src/test/java/com/segment/analytics/kotlin/android/AndroidLifecyclePluginTests.kt
```diff
@@ -364,7 +345,7 @@ class AndroidLifecyclePluginTests {
         // Simulate activity startup
         lifecyclePlugin.onActivityCreated(mockActivity, mockBundle)
 
-        assertTrue(mockPlugin.ran)
+        verify (timeout = 2000){  mockPlugin.updateState(true) }
```
prayansh, line 348: is this timeout necessary due to us not being able to run the tests with the `testDispatcher`?

prayansh (approved):

android/src/test/java/com/segment/analytics/kotlin/android/AndroidLifecyclePluginTests.kt
wenxi-zeng: correct. because we are not able to use `testDispatcher`, things running async. there is a delay before the update happen. the `verify` waits until the statement happened or timeout reached, whichever one comes first.

core/src/main/java/com/segment/analytics/kotlin/core/Configuration.kt
```diff
@@ -47,6 +43,13 @@ data class Configuration(
     var apiHost: String = DEFAULT_API_HOST,
     var cdnHost: String = DEFAULT_CDN_HOST
 ) {
+    internal var analyticsDispatcher: CoroutineDispatcher = Dispatchers.IO
+
+    internal var ioDispatcher: CoroutineDispatcher = Executors.newFixedThreadPool(1)
+        .asCoroutineDispatcher()
+
+    internal var analyticsScope: CoroutineScope = MainScope()
```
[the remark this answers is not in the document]
wenxi-zeng: agree. I was also worried about it running in main thread. changing it now

PR #38, opened 2021-09-15
increase test coverage

wenxi-zeng:
* increase test coverage from class 85% line 84% to class 95% line 94%
* refactor test file corresponding folder

### DIFF
--- a/core/src/main/java/com/segment/analytics/kotlin/core/HTTPClient.kt
+++ b/core/src/main/java/com/segment/analytics/kotlin/core/HTTPClient.kt
@@ -78,11 +78,12 @@ abstract class Connection(
     }
 }
 
-fun safeGetInputStream(connection: HttpURLConnection): InputStream {
+fun safeGetInputStream(connection: HttpURLConnection): InputStream? {
     return try {
         connection.inputStream
     } catch (ignored: IOException) {
-        connection.errorStream
+        // cannot use connection.errorStream here, since it returns null which breaks non-null check in kotlin
+        null
     }
 }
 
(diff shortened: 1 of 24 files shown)


core/src/main/java/com/segment/analytics/kotlin/core/HTTPClient.kt
```diff
@@ -78,11 +78,12 @@ abstract class Connection(
     }
 }
 
-fun safeGetInputStream(connection: HttpURLConnection): InputStream {
+fun safeGetInputStream(connection: HttpURLConnection): InputStream? {
     return try {
         connection.inputStream
     } catch (ignored: IOException) {
-        connection.errorStream
```
prayansh, line 85: im a bit confused on this change

prayansh (approved):

core/src/main/java/com/segment/analytics/kotlin/core/HTTPClient.kt
wenxi-zeng: ah. yeah. this one is a weird one. if you check the source of `connection.errorStream`, it's actually returning null.
```java
    public InputStream getErrorStream() {
        return null;
    }
```
so we need to change the return type to `InputStream?`, or it will throws an exception and crashes the sdk. also since `connection.stream` always returning null, we can just return null in our code as well. but returning `connection.errorStream` also works and has better readability

prayansh: lets keep it as `connection.errorStream`. The default implementation might return `null` but i believe that method can be overridden by other implementations which may not return `null`.

wenxi-zeng: yeah. good point. that can totally happen. making the change now

PR #39, opened 2021-09-16
Fix destination processing logic

prayansh:
This PR fixes Destination plugins not propagating changes made by the destination to the `After` plugins


wenxi-zeng (approved):
LGTM

PR #43, opened 2021-09-24
Mixpanel + Firebase Refactor

prayansh:
This PR refactors the Mixpanel and Firebase destination samples to utilise the new library APIs

I also ended up changing the circle config to be more modular for testing


samples/kotlin-android-app-destinations/src/main/java/com/segment/analytics/destinations/plugins/MixpanelDestination.kt
```diff
@@ -4,142 +4,181 @@ import android.app.Activity
...
-    private var trackAllPages = false
-    private var trackCategorizedPages = false
-    private var trackNamedPages = false
-    private var peoplePropertiesFilter: Set<String> = setOf("peopleProperties")
-    private var superPropertiesFilter: Set<String> = setOf("superProperties")
-    private var increments: Set<String> = setOf("superProperties")
-
-    override fun track(payload: TrackEvent): BaseEvent {
+    override fun track(payload: TrackEvent): BaseEvent? {
+        val settings = settings ?: return null
```
wenxi-zeng, line 93: can we return `payload` instead of `null` here?

samples/kotlin-android-app-destinations/src/main/java/com/segment/analytics/destinations/plugins/MixpanelDestination.kt
```diff
@@ -4,142 +4,181 @@ import android.app.Activity
...
-        analytics.log("mixpanel.identify($userId)", type = LogType.INFO, event = payload)
-
-        // Filter based on configuration and transform keys using PROPERTY_MAPPER
-        val peopleProperties = if (setAllTraitsByDefault || peoplePropertiesFilter.isEmpty()) {
-            payload.traits
-        } else {
-            payload.traits.filter { (k, _) -> peoplePropertiesFilter.contains(k) }
-        }.mapKeys { TRAITS_MAPPER[it.key] ?: it.key }
+    override fun identify(payload: IdentifyEvent): BaseEvent? {
+        val settings = settings ?: return null
```
wenxi-zeng, line 111: same question here

samples/kotlin-android-app-destinations/src/test/java/com/segment/analytics/destinations/plugins/FirebaseDestinationTest.kt
```diff
@@ -0,0 +1,304 @@
...
+        """.trimIndent()
+        )
+        firebaseDestination.update(settingsBlob, Plugin.UpdateType.Initial)
+
+        assertNotNull(firebaseDestination.firebaseAnalytics)
+    }
+
+    @Test
+    fun `identify is handled correctly`() {
+        firebaseDestination.firebaseAnalytics = FirebaseAnalytics.getInstance(context)
```
wenxi-zeng, line 77: not sure if the following works for this case, but we can mock private properties by doing something like this
```kotlin
firebaseDestination = spyk(FirebaseDestination())
every { firebaseDestination getProperty "firebaseAnalytics" } propertyType FirebaseAnalytics::class returns mockedFirebase
```
in this case, we can keep `firebaseAnalytics` private in `FirebaseDestination`, and move the setup the `setup` method avoiding doing it for every unit test

samples/kotlin-android-app-destinations/src/test/java/com/segment/analytics/destinations/plugins/MixpanelDestinationTests.kt
```diff
@@ -0,0 +1,902 @@
...
+        }
+
+        // groupKey defaults to groupId, since no name is set
+        verify { mockMixpanel.setGroup("ABC network", "grp-123") }
+    }
+
+    /* PRIVATE TEST FUNCTIONS */
+
+    private fun MixpanelDestination.setupTest(mixpanelSettings: MixpanelSettings) {
+        this.mixpanel = mockMixpanel
```
wenxi-zeng, line 879: same as this one. probably can use the way of mock private properties for setup, and keep these properties as private

samples/kotlin-android-app-destinations/src/test/java/com/segment/analytics/destinations/plugins/FirebaseDestinationTest.kt
prayansh: we definitely do not want to mock / spy on the main testing object i.e. `firebaseDestination`. 

The only reason i am doing this here is cos technically that initialization during the `update` and i dont want to call update everytime i do a test, so this is more like a test setup thing. Also this needs to happen for all but settings tests... Maybe there is a better way to do this 🤔

wenxi-zeng: yeah. I agree we should not mock the main testing object. but I think spy in some cases are fine, like for this case, it just injects a mocked property. other behaviors are still the actual behaviors. but you're also right, that's kind risky if we're not paying attention. I think the current way is fine, it's just a bit strange that we makes properties `internal` just because we need them in unit tests.  😅

wenxi-zeng (approved):

PR #44, opened 2021-09-28
Add sample appsflyer destination plugin

prayansh:


samples/kotlin-android-app-destinations/src/main/java/com/segment/analytics/destinations/plugins/AppsflyerDestination.kt
```diff
@@ -0,0 +1,244 @@
...
+
+@Serializable
+data class AppsFlyerSettings(
+    var appsFlyerDevKey: String,
+    var trackAttributionData: Boolean = false
+)
+
+class AppsFlyerDestination(
+    private val applicationContext: Context,
+    private var isDebug: Boolean = false
```
wenxi-zeng, line 64: maybe we can check if `analytics` has `logger` plugin to determine `isDebug`?

samples/kotlin-android-app-destinations/src/main/java/com/segment/analytics/destinations/plugins/AppsflyerDestination.kt
```diff
@@ -0,0 +1,244 @@
...
+                else -> JsonPrimitive(value.toString())
+            }
+        }
+
+        private fun trackInstallAttributed(attributionData: Map<String, Any>) {
+            // See https://segment.com/docs/spec/mobile/#install-attributed.
+            val properties = buildJsonObject {
+                put("provider", key)
+                attributionData.forEach { (k, v) ->
+                    if (k != "media_source" && k != "adGroup") {
```
wenxi-zeng, line 210: nit: maybe `k !in setOf("media_source", "adGroup")` is better? in case we are excluding other keys in the future.

samples/kotlin-android-app-destinations/src/main/java/com/segment/analytics/destinations/plugins/AppsflyerDestination.kt
```diff
@@ -0,0 +1,244 @@
...
+            }
+
+            // If you are working with networks that don't allow passing user level data to 3rd parties,
+            // you will need to apply code to filter out these networks before calling
+            // `analytics.track("Install Attributed", properties);`
+            analytics.track("Install Attributed", properties)
+        }
+
+        private fun getFlag(key: String): Boolean {
+            val context: Context = applicationContext
```
wenxi-zeng, line 228: nit: this line seems redundant. we can directly use `applicationConext`

samples/kotlin-android-app-destinations/src/main/java/com/segment/analytics/destinations/plugins/AppsflyerDestination.kt
```diff
@@ -0,0 +1,244 @@
...
+            val sharedPreferences: SharedPreferences =
+                context.getSharedPreferences(AF_SEGMENT_SHARED_PREF, 0)
+            return sharedPreferences.getBoolean(key, false)
+        }
+
+        private fun setFlag(key: String, value: Boolean) {
+            val context: Context = applicationContext
```
wenxi-zeng, line 235: same here

wenxi-zeng (commented):
is it possible to also test `ConversionListener`?

samples/kotlin-android-app-destinations/src/main/java/com/segment/analytics/destinations/plugins/AppsflyerDestination.kt
prayansh: the logger plugin is added to analytics by default, so this would result in it being always true. 

I think providing the user the option to specify when they want `debug` to be `true/false` directly is better than to derive that value via some other property. I could be convinced otherwise, but i would rather keep this

samples/kotlin-android-app-destinations/src/main/java/com/segment/analytics/destinations/plugins/AppsflyerDestination.kt
prayansh: i like that. will change

samples/kotlin-android-app-destinations/src/main/java/com/segment/analytics/destinations/plugins/AppsflyerDestination.kt
wenxi-zeng: 👍 yeah. that makes sense. I also felt that's not a reliable way.

wenxi-zeng (approved):
LGTM

PR #52, opened 2021-10-12
setup github actions

wenxi-zeng:
* setup github actions
* remove circleci
* update libraries version
* add snyk test
* fix flaky unit tests

### DIFF
--- a/android/src/test/java/com/segment/analytics/kotlin/android/AndroidLifecyclePluginTests.kt
+++ b/android/src/test/java/com/segment/analytics/kotlin/android/AndroidLifecyclePluginTests.kt
@@ -13,12 +13,9 @@ import com.segment.analytics.kotlin.core.platform.Plugin
 import com.segment.analytics.kotlin.android.plugins.AndroidLifecycle
 import com.segment.analytics.kotlin.android.plugins.AndroidLifecyclePlugin
 import com.segment.analytics.kotlin.android.utils.mockHTTPClient
-import com.segment.analytics.kotlin.android.utils.spyStore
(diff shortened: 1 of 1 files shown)


android/src/test/java/com/segment/analytics/kotlin/android/AndroidLifecyclePluginTests.kt
```diff
@@ -54,13 +62,19 @@ class AndroidLifecyclePluginTests {
...
                 storageProvider = AndroidStorageProvider
             )
-        )
+        ))
+        every { analytics.analyticsScope } returns testScope
+        every { analytics.analyticsDispatcher } returns testDispatcher
+        every { analytics.networkIODispatcher } returns testDispatcher
+        every { analytics.fileIODispatcher } returns testDispatcher
+        every { analytics.store } returns spyStore(testScope, testDispatcher)
+        lifecyclePlugin = AndroidLifecyclePlugin()
```
prayansh, line 77: any reason why this needs to be initialized here vs how it was before

prayansh (approved):

android/src/test/java/com/segment/analytics/kotlin/android/AndroidLifecyclePluginTests.kt
wenxi-zeng: yeah. this does not have to be here. I will revert this change. I was blindly trying to fix the flaky test. this eliminated a possible cause, since `AndroidLifecyclePlugin` manages some states internally.

PR #53, opened 2021-11-01
Wenxi/replace flush upload with channel api

wenxi-zeng:
* minimize the number of times of open/close file
* update `SegmentDestination` with a pipeline implementation using channel api
* make `analytics.flush` thread safe
* make file writing thread safe
* fix a bug when fetch settings
* update unit tests

### DIFF
--- a/android/src/main/java/com/segment/analytics/kotlin/android/Storage.kt
+++ b/android/src/main/java/com/segment/analytics/kotlin/android/Storage.kt
@@ -46,7 +46,7 @@ class AndroidStorage(
         )
     }
 
-    override fun write(key: Storage.Constants, value: String) {
+    override suspend fun write(key: Storage.Constants, value: String) {
         when (key) {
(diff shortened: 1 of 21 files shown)


android/src/main/java/com/segment/analytics/kotlin/android/Storage.kt
```diff
@@ -92,6 +92,10 @@ class AndroidStorage(
     override fun removeFile(filePath: String): Boolean {
         return eventsFile.remove(filePath)
     }
+
+    override suspend fun rollover() {
```
prayansh, line 96: can we add some docs around what this function should be doing

prayansh (approved):

android/src/main/java/com/segment/analytics/kotlin/android/Storage.kt
wenxi-zeng: good call. will do

PR #54, opened 2021-11-02
add cancel previous

wenxi-zeng:
* add ability to cancel previous job

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,17 +8,8 @@ on:
   workflow_dispatch:
 
 jobs:
-  cancel_previous:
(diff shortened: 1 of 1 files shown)


prayansh (approved):

prayansh:
why do we wanna distinguish cache names?

wenxi-zeng:
> why do we wanna distinguish cache names?

good question. I think the VMs should be smart enough to have different hash for different jobs (core, android, dest), since their dependencies are different. it's just in case they generate the same hash for other jobs. in that case, for example, `core` job ends up using the cache of `android`.

PR #56, opened 2021-11-09
Migs647/logger

migs647:
Added new API logger


core/src/main/java/com/segment/analytics/kotlin/core/Analytics.kt
```diff
@@ -40,6 +41,12 @@ class Analytics internal constructor(
 
     internal val timeline: Timeline
     val storage: Storage
+    companion object {
+        var debugLogsEnabled: Boolean = false
```
prayansh, line 45: whats the use-case of this variable?

core/src/main/java/com/segment/analytics/kotlin/core/platform/plugins/StartupQueue.kt
```diff
@@ -40,7 +41,7 @@ class StartupQueue : Plugin, Subscriber {
 
     override fun execute(event: BaseEvent): BaseEvent? {
         if (!started.get()) {
-            analytics.log("SegmentStartupQueue queueing event", event = event)
+            analytics.log("SegmentStartupQueue queueing event")
```
prayansh, line 44: why not log the event?

android/src/main/java/com/segment/analytics/kotlin/android/AndroidAnalytics.kt
```diff
@@ -54,28 +51,8 @@ public fun Analytics(
     }
 }
 
-// Logger instance that uses the android `Log` class
-object AndroidLogger : Logger() {
```
prayansh, line 58: since we are deleting this. we should add a new AndroidLogTarget

prayansh (commented):
couple comments

core/src/main/java/com/segment/analytics/kotlin/core/platform/plugins/logger/SegmentLog.kt
```diff
@@ -0,0 +1,187 @@
...
+                             override val function: String?,
+                             override val line: Int?,
+                             val sender: Any?,
+                             override val logType: LoggingType.Filter = LoggingType.Filter.HISTORY,
+                             override val dateTime: Date = Date()
+    ): LogMessage
+}
+
+// Default implementation
+fun LogTarget.flush() {
```
prayansh, line 132: this feels like a swift thing but not a kotlin thing. normally u would just make the interface have a default empty implementation (https://github.com/segmentio/analytics-kotlin/pull/56/files#diff-de76397a2e004dcb8d9f3e1537111c83c5adc5148ee82a830dced7538b524efdR27)

core/src/main/java/com/segment/analytics/kotlin/core/Analytics.kt
migs647: This is how outside engineers can override the switch from the web.

core/src/main/java/com/segment/analytics/kotlin/core/platform/plugins/StartupQueue.kt
migs647: analytics.log is all about a simple log message. Logging the event would use the history option. Event is a ton of metadata when we're just trying to capture certain elements they are free to pass in.

core/src/main/java/com/segment/analytics/kotlin/core/platform/plugins/logger/SegmentLog.kt
migs647: Ahh that's what i was going for here. I'll refactor.

core/src/main/java/com/segment/analytics/kotlin/core/platform/plugins/logger/LogTarget.kt
```diff
@@ -0,0 +1,238 @@
...
+        }
+    }
+}
+
+/**
+ * The Segment logging system has three types of logs: log, metric and history. When adding a target that
+ * responds to logs, it is possible to adhere to 1 to many. In other words, a LoggingType can be .log &
+ * .history. This is used to tell which targets logs are directed to.
+ */
+class LoggingType(types: List<Filter>) {
```
wenxi-zeng, line 52: can do it the following way. then line 70 can be removed.
```suggestion
class LoggingType(private val allTypes: List<Filter>) {
```

core/src/main/java/com/segment/analytics/kotlin/core/platform/plugins/logger/LogTarget.kt
```diff
@@ -0,0 +1,238 @@
...
+    // - Private Properties and Methods
+    private val allTypes: List<Filter> = types
+
+    /**
+     * Convenience method to find if the LoggingType supports a particular destination.
+     *
+     * @property destination The particular destination being tested for conformance.
+     */
+    internal fun contains(destination: Filter): Boolean {
+        return this.allTypes.contains(destination)
```
wenxi-zeng, line 78: does `allTypes` allow duplicate filters/destinations? if not, it's probably better to use `Set` instead of `List` for better performance.

core/src/main/java/com/segment/analytics/kotlin/core/platform/plugins/logger/LogTarget.kt
```diff
@@ -0,0 +1,238 @@
...
+    val title: String?
+    val message: String
+    val event: BaseEvent?
+    val function: String?
+    val line: Int?
+    val logType: LoggingType.Filter
+    val dateTime: Date
+}
+
+enum class MetricType(val type: Int) {
```
wenxi-zeng, line 96: is `type` necessary here? enum already has a `ordinal` property that seems to serve the same purpose

core/src/main/java/com/segment/analytics/kotlin/core/platform/plugins/logger/LogTarget.kt
```diff
@@ -0,0 +1,238 @@
...
+    Counter(0), // Not Verbose
+    Gauge(1);   // Semi-verbose
+
+    override fun toString(): String {
```
wenxi-zeng, line 100: can be simplified
```suggestion
    override fun toString() = this.name
```

core/src/main/java/com/segment/analytics/kotlin/core/platform/plugins/logger/LogTarget.kt
```diff
@@ -0,0 +1,238 @@
...
+        if (this == Counter) {
+            typeString = "Counter"
+        }
+        return typeString
+    }
+
+    companion object {
+        fun from(string: String): MetricType {
+            var returnType = Counter
+            if (string == "Gauge") {
```
wenxi-zeng, line 111: we can use the following to avoid using hardcoded value
```suggestion
            if (valueOf(string) == Gauge) {
```

core/src/main/java/com/segment/analytics/kotlin/core/platform/plugins/logger/LogTarget.kt
```diff
@@ -0,0 +1,238 @@
...
+ * @property kind Usually .error, .warning or .debug, in order of severity. This helps filter logs based on
+ * this added metadata.
+ * @property function The name of the function the log came from. This will be captured automatically.
+ * @property line The line number in the function the log came from. This will be captured automatically.
+ */
+@JvmOverloads
+fun Analytics.log(message: String, kind: LogFilterKind? = null, function: String = "", line: Int = -1) {
+    applyClosureToPlugins { plugin: Plugin ->
+        // Check if we should send the event
+        if (!SegmentLog.loggingEnabled) {
```
wenxi-zeng, line 133: it's probably better to do this check before line 131 `applyClosureToPlugins`, so that we have an earlier termination.  otherwise, it goes through unnecessary loops that iterates the plugins.

core/src/main/java/com/segment/analytics/kotlin/core/platform/plugins/logger/LogTarget.kt
```diff
@@ -0,0 +1,238 @@
...
+ * @property name The title of the metric to track.
+ * @property value The value associated with the metric. This would be an incrementing counter or time
+ * or pressure gauge.
+ * @property tags Any tags that should be associated with the metric. Any extra metadata that may help.
+ */
+@JvmOverloads
+fun Analytics.metric(type: String, name: String, value: Double, tags: List<String>? = null) {
+    applyClosureToPlugins { plugin: Plugin ->
+        // Check if we should send the event
+        if (!SegmentLog.loggingEnabled) {
```
wenxi-zeng, line 162: same thing here

core/src/main/java/com/segment/analytics/kotlin/core/platform/plugins/logger/LogTarget.kt
```diff
@@ -0,0 +1,238 @@
...
+    val methodInfo = Analytics.callingMethodDetails(function, line)
+    var updatedFunction = function
+    var updatedLine = line
+    if (function.isEmpty() && line == -1) {
+        updatedFunction = methodInfo.first
+        updatedLine = methodInfo.second
+    }
+    applyClosureToPlugins { plugin: Plugin ->
+        // Check if we should send the event
+        if (!SegmentLog.loggingEnabled) {
```
wenxi-zeng, line 195: and here

core/src/main/java/com/segment/analytics/kotlin/core/platform/plugins/logger/LogTarget.kt
migs647: Good call, not just better performance but better api

core/src/main/java/com/segment/analytics/kotlin/core/platform/plugins/logger/LogTarget.kt
migs647: Changed

core/src/main/java/com/segment/analytics/kotlin/core/platform/plugins/logger/LogTarget.kt
migs647: Yep good catch

wenxi-zeng:
👍  changes LGTM. just need to fix the android and destination tests now

android/src/main/java/com/segment/analytics/kotlin/android/AndroidAnalytics.kt
```diff
@@ -55,4 +56,36 @@ public fun Analytics(
...
+
+        when (log.kind) {
+            LogFilterKind.ERROR -> {
+                Log.e("AndroidLog", "message=${log.message}$eventString")
+            }
+            LogFilterKind.WARNING -> {
+                Log.w("AndroidLog", "message=${log.message}$eventString")
+            }
+            LogFilterKind.DEBUG -> {
+                Log.i("AndroidLog", "message=${log.message}$eventString")
```
prayansh, line 85: any reason we are using `i` instead of `d` here?

wenxi-zeng (approved):

PR #57, opened 2021-11-17
Wenxi/fix identity doc

wenxi-zeng:
* fix identity doc where it says userid can be null
* add reset function for user to call after logged out
* add unit tests

addition:
* fix `find` discrepancy behavior
* add `findAll`
* optimize segment log performance a bit
* add unit test for `find` and `findAll`


core/src/main/java/com/segment/analytics/kotlin/core/platform/Mediator.kt
```diff
@@ -65,10 +65,14 @@ internal class Mediator(internal val plugins: MutableList<Plugin>) {
 
     fun <T: Plugin> find(pluginClass: KClass<T>): T? = synchronized(plugins) {
         plugins.forEach {
-            if (it::class == pluginClass) {
+            if (pluginClass.java.isAssignableFrom(it::class.java)) {
```
prayansh, line 68: hmmm is this the most idiomatic way? what are we trying to achieve here?

prayansh (approved):

PR #59, opened 2021-11-29
Integrations Logic fix + Copy mechanism for payloads

prayansh:
- Correct how events select which device-mode destinations to run against
- Move logic for device-mode destinations data to Segment.io destination plugin
- Add a new API to allow manually enabling destinations (useful for destination plugins with no Segment tie-ins)
- Add a new API to allow copying payloads with ease

### DIFF
--- a/core/src/main/java/com/segment/analytics/kotlin/core/Analytics.kt
+++ b/core/src/main/java/com/segment/analytics/kotlin/core/Analytics.kt
@@ -382,11 +382,6 @@ class Analytics internal constructor(
      */
     fun add(plugin: Plugin): Analytics {
         this.timeline.add(plugin)
-        if (plugin is DestinationPlugin && plugin !is SegmentDestination) {
-            analyticsScope.launch(analyticsDispatcher) {
-                store.dispatch(System.AddIntegrationAction(plugin.key), System::class)
-            }
-        }
         return this
     }
 
@@ -412,11 +407,6 @@ class Analytics internal constructor(
      */
     fun remove(plugin: Plugin): Analytics {
         this.timeline.remove(plugin)
-        if (plugin is DestinationPlugin && plugin !is SegmentDestination) {
-            analyticsScope.launch(analyticsDispatcher) {
-                store.dispatch(System.RemoveIntegrationAction(plugin.key), System::class)
-            }
-        }
         return this
     }
 
@@ -502,6 +492,24 @@ class Analytics internal constructor(
         return traitsAsync()?.let {
             decodeFromJsonElement(deserializationStrategy, it)
         }
+    }
+
+    /**
+     * Retrieve the settings  in a blocking way.
+     * Note: this method invokes `runBlocking` internal, it's not recommended to be used
+     * in coroutines.
+     */
+    @BlockingApi
+    fun settings(): Settings? = runBlocking {
+        settingsAsync()
+    }
+
+    /**
+     * Retrieve the settings
+     */
+    suspend fun settingsAsync(): Settings? {
+        val system = store.currentState(System::class)
+        return system?.settings
     }
 }
 
--- a/core/src/main/java/com/segment/analytics/kotlin/core/Events.kt
+++ b/core/src/main/java/com/segment/analytics/kotlin/core/Events.kt
@@ -90,7 +90,7 @@ sealed class BaseEvent {
         val userInfo = store.currentState(UserInfo::class) ?: return
 
         this.anonymousId = userInfo.anonymousId
-        this.integrations = system.integrations ?: emptyJsonObject
+        this.integrations = emptyJsonObject
 
         if (this.userId.isBlank()) {
             // attach system userId if present
--- a/core/src/main/java/com/segment/analytics/kotlin/core/Events.kt
+++ b/core/src/main/java/com/segment/analytics/kotlin/core/Events.kt
@@ -86,7 +86,6 @@ sealed class BaseEvent {
     }
 
     internal suspend fun applyBaseEventData(store: Store) {
-        val system = store.currentState(System::class) ?: return
         val userInfo = store.currentState(UserInfo::class) ?: return
 
         this.anonymousId = userInfo.anonymousId
@@ -96,6 +95,60 @@ sealed class BaseEvent {
             // attach system userId if present
             this.userId = userInfo.userId ?: ""
         }
+    }
+
+    // Create a shallow copy of this event payload
+    fun <T : BaseEvent> copy(): T {
+        val original = this
+        val copy = when (this) {
+            is AliasEvent -> AliasEvent(userId = this.userId, previousId = this.previousId)
+            is GroupEvent -> GroupEvent(groupId = this.groupId, traits = this.traits)
+            is IdentifyEvent -> IdentifyEvent(userId = this.userId, traits = this.traits)
+            is ScreenEvent -> ScreenEvent(
+                name = this.name,
+                category = this.category,
+                properties = this.properties
+            )
+            is TrackEvent -> TrackEvent(event = this.event, properties = this.properties)
+        }.apply {
+//            type = original.type
+            anonymousId = original.anonymousId
+            messageId = original.messageId
+            timestamp = original.timestamp
+            context = original.context
+            integrations = original.integrations
+            userId = original.userId
+        }
+        @Suppress("UNCHECKED_CAST")
+        return copy as T // This is ok because resultant type will be same as input type
+    }
+
+    override fun equals(other: Any?): Boolean {
+        if (this === other) return true
+        if (javaClass != other?.javaClass) return false
+
+        other as BaseEvent
+
+        if (type != other.type) return false
+        if (anonymousId != other.anonymousId) return false
+        if (messageId != other.messageId) return false
+        if (timestamp != other.timestamp) return false
+        if (context != other.context) return false
+        if (integrations != other.integrations) return false
+        if (userId != other.userId) return false
+
+        return true
+    }
+
+    override fun hashCode(): Int {
+        var result = type.hashCode()
+        result = 31 * result + anonymousId.hashCode()
+        result = 31 * result + messageId.hashCode()
+        result = 31 * result + timestamp.hashCode()
+        result = 31 * result + context.hashCode()
+        result = 31 * result + integrations.hashCode()
+        result = 31 * result + userId.hashCode()
+        return result
     }
 }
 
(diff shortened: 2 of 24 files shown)


core/src/main/java/com/segment/analytics/kotlin/core/Settings.kt
```diff
@@ -54,6 +74,7 @@ suspend fun Analytics.checkSettings() {
 
     // check current system state to determine whether it's initial or refresh
     val systemState = store.currentState(System::class)
+    println(systemState?.settings)
```
wenxi-zeng, line 77: seems to be a debugging output. should remvoe

core/src/main/java/com/segment/analytics/kotlin/core/platform/Plugin.kt
```diff
@@ -136,4 +137,15 @@ abstract class DestinationPlugin : EventPlugin {
     final override fun execute(event: BaseEvent): BaseEvent? {
         return null
     }
+
+    internal fun isDestinationEnabled(event: BaseEvent?): Boolean {
+        // if event payload has integration marked false then its disabled by customer
+        val customerDisabled = when (event?.integrations?.getBoolean(key)) {
```
wenxi-zeng, line 143: how about the following?
```suggestion
        val customerEnable = event?.integrations?.getBoolean(key) ?: true
```
avoiding double negation for better readability, so when return 
```
    return (enabled && customerEnable)
```

core/src/main/java/com/segment/analytics/kotlin/core/Settings.kt
```diff
@@ -29,22 +30,41 @@ data class Settings(
...
         // tell all top level plugins to update.
         // For destination plugins they auto-handle propagation to sub-plugins
         plugin.update(settings, type)
     }
 }
 
+fun Analytics.manuallyEnableDestination(plugin: DestinationPlugin) {
+    analyticsScope.launch(analyticsDispatcher) {
+        store.dispatch(
+            System.AddDestinationToSettingsAction(destinationKey = plugin.key) { settings: Settings? ->
```
wenxi-zeng, line 56: probably don't need the `afterClosure`, since `dispatch` is a suspend function, we can move the `if` block below `store.dispacth`

prayansh: we still need the after closure bcos we need the resultant `settings` from the action

prayansh: do u think this is a better way to do it?
```kotlin
fun Analytics.manuallyEnableDestination(plugin: DestinationPlugin) {
    analyticsScope.launch(analyticsDispatcher) {
        store.dispatch(
            System.AddDestinationToSettingsAction(destinationKey = plugin.key),
            System::class
        )
        val system = store.currentState(System::class)
        system?.settings?.let { settings ->
            findAll(DestinationPlugin::class).forEach {
                plugin.enabled = settings.hasIntegrationSettings(it.key)
            }
        }
    }
}
```

wenxi-zeng: yeah. I think this is better, though performance-wise it's a bit slower. one of the benefits of suspend function is avoiding indentation/callback hell. I think we should take it.

core/src/main/java/com/segment/analytics/kotlin/core/Events.kt
```diff
@@ -86,17 +86,70 @@ sealed class BaseEvent {
...
+        this.integrations = emptyJsonObject
 
         if (this.userId.isBlank()) {
             // attach system userId if present
             this.userId = userInfo.userId ?: ""
         }
     }
+
+    // Create a shallow copy of this event payload
+    fun <T : BaseEvent> copy(): T {
```
wenxi-zeng, line 101: can we make this method abstract, let the subclasses implement the copy details? that case, the parent class doesn't have to know it's children. if we add more subclasses, we don't need to modify the parent class.

core/src/main/java/com/segment/analytics/kotlin/core/Events.kt
```diff
@@ -86,17 +86,70 @@ sealed class BaseEvent {
...
+
+        if (type != other.type) return false
+        if (anonymousId != other.anonymousId) return false
+        if (messageId != other.messageId) return false
+        if (timestamp != other.timestamp) return false
+        if (context != other.context) return false
+        if (integrations != other.integrations) return false
+        if (userId != other.userId) return false
+
+        return true
```
wenxi-zeng, line 140: not necessary better, but can be simplified as 
```suggestion
        return type == other.type &&
                    anonymousId == other.anonymousId &&
                    messageId == other.messageId &&
                    timestamp == other.timestamp &&
                    context == other.context &&
                    integrations == other.integrations &&
                    userId != other.userId
```

core/src/main/java/com/segment/analytics/kotlin/core/Events.kt
```diff
@@ -113,6 +166,27 @@ data class TrackEvent(
...
+        if (this === other) return true
+        if (javaClass != other?.javaClass) return false
+        if (!super.equals(other)) return false
+
+        other as TrackEvent
+
+        if (properties != other.properties) return false
+        if (event != other.event) return false
+
+        return true
```
wenxi-zeng, line 180: same here

core/src/main/java/com/segment/analytics/kotlin/core/platform/plugins/SegmentDestination.kt
```diff
@@ -99,4 +103,20 @@ class SegmentDestination : DestinationPlugin() {
     override fun flush() {
         pipeline.flush()
     }
+
+    internal fun <T: BaseEvent> configureCloudDestination(event: T): T {
+        val enabledDestinations = analytics.findAll(DestinationPlugin::class).filter { it.enabled }
```
wenxi-zeng, line 108: doing a `findAll` for each of the enqueuing event could be a bottle neck, though the total number of plugins could be small, it does not scale well.
is it possible to make `SegmentDestination` a `Subscriber` of `System` state? we cache the `enabledDestinations`, and update it when there is plugin changes. 
just a suggestion, may not be feasible 😄

core/src/main/java/com/segment/analytics/kotlin/core/Events.kt
prayansh: i chose to do this in the BaseEvent class for a 2 reasons
- We know that we wont have more than track/identify/group/screen/alias event types
- BaseEvent is a sealed type so its ok to have the full implementation in the parent class (which knows about all the subclasses)

core/src/main/java/com/segment/analytics/kotlin/core/Events.kt
prayansh: i could be wrong, but isn't the current one better for debuggability? if not then I will change it to your suggestion

core/src/main/java/com/segment/analytics/kotlin/core/platform/plugins/SegmentDestination.kt
prayansh: lemme try that

core/src/main/java/com/segment/analytics/kotlin/core/Events.kt
wenxi-zeng: that makes sense

core/src/main/java/com/segment/analytics/kotlin/core/Events.kt
wenxi-zeng: yeah, the current one is actually better for debugability. no need to change it. I believe the compiler will optimize it itself. I'm still live in fear of java word where the compiler warns about this can be simplified 😄.

wenxi-zeng: yeah. the current one is better for deubgability. no need to change it. I believe the compiler will optimize it itself. I'm still live in the fear of java word where the compiler warns this can be simplified. 😄

core/src/main/java/com/segment/analytics/kotlin/core/platform/plugins/SegmentDestination.kt
prayansh: i went with a different route here so as to make it similar to the swift code

core/src/main/java/com/segment/analytics/kotlin/core/Analytics.kt
```diff
@@ -503,6 +493,24 @@ class Analytics internal constructor(
...
         }
     }
+
+    /**
+     * Retrieve the settings  in a blocking way.
+     * Note: this method invokes `runBlocking` internal, it's not recommended to be used
+     * in coroutines.
+     */
+    @BlockingApi
+    fun settings(): Settings? = runBlocking {
```
wenxi-zeng, line 503: need to add a `settings` function to `JavaAnalytics` for compatibility. also unit tests for them

core/src/main/java/com/segment/analytics/kotlin/core/Settings.kt
```diff
@@ -79,6 +104,7 @@ suspend fun Analytics.checkSettings() {
                 log("Dispatching update settings on ${Thread.currentThread().name}")
                 store.dispatch(System.UpdateSettingsAction(settingsObj), System::class)
                 update(settingsObj, updateType)
+                store.dispatch(System.ToggleSettingsDispatch(dispatched = true), System::class)
```
wenxi-zeng, line 107: just to confirm: we dispatch twice here because we want to ensure settings being successfully propagated to subscribers and then toggle the state?

wenxi-zeng (approved):
LGTM

core/src/main/java/com/segment/analytics/kotlin/core/Settings.kt
prayansh: yup i didnt want to preemptively toggle the dispatch

core/src/main/java/com/segment/analytics/kotlin/core/Analytics.kt
prayansh: ah good catch. will add those

core/src/main/java/com/segment/analytics/kotlin/core/Settings.kt
```diff
@@ -50,18 +52,19 @@ internal fun Analytics.update(settings: Settings, type: Plugin.UpdateType) {
...
         )
-        val system = store.currentState(System::class)
-        system?.settings?.let { settings ->
-            findAll(DestinationPlugin::class).forEach {
-                plugin.enabled = settings.hasIntegrationSettings(it.key)
-            }
-        }
+        // Differs from swift, bcos kotlin can store `enabled` state. ref: https://git.io/J1bhJ
+        // finding it in timeline rather than using the ref that is provided to cover our bases
+        find(plugin::class)?.enabled = true
```
wenxi-zeng, line 67: ❤️ nice and clean

PR #60, opened 2021-11-30
Wenxi/convenience methods and misses for plugin authors

wenxi-zeng:
* added `enableIntegration`, `disableIntegration`, `disableCloudIntegrations`, `enableCloudIntegrations`
* added `updateJsonObject` to JSON wrapper for easier content manipulation
* added unit tests
* fixed a bug in intercom integration
* `manuallyEnableDestination` is already implemented in pr #59

### DIFF
--- a/core/src/main/java/com/segment/analytics/kotlin/core/utilities/JSON.kt
+++ b/core/src/main/java/com/segment/analytics/kotlin/core/utilities/JSON.kt
@@ -181,3 +181,27 @@ fun JsonObject.transformKeys(transform: (String) -> String): JsonObject {
 fun JsonObject.transformValues(transform: (JsonElement) -> JsonElement): JsonObject {
     return JsonObject(this.mapValues { transform(it.value) })
 }
+
+// Utility function to update a JsonObject. The updated copy is returned. Original copy is not touched
+fun updateJsonObject(jsonObject: JsonObject, closure: (MutableMap<String, JsonElement>) -> Unit): JsonObject {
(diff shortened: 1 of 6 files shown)


core/src/main/java/com/segment/analytics/kotlin/core/utilities/JSON.kt
```diff
@@ -181,3 +181,27 @@ fun JsonObject.transformKeys(transform: (String) -> String): JsonObject {
...
 }
+
+// Utility function to update a JsonObject. The updated copy is returned. Original copy is not touched
+fun updateJsonObject(jsonObject: JsonObject, closure: (MutableMap<String, JsonElement>) -> Unit): JsonObject {
+    val content = jsonObject.toMutableMap()
+    closure(content)
+    return JsonObject(content)
+}
+
+operator fun MutableMap<String, JsonElement>.set(key:String, value: String?) {
```
prayansh, line 192: would adding these APIs on such a widely used type (like MutableMap) have consequences?

prayansh (approved):
looks good

core/src/main/java/com/segment/analytics/kotlin/core/utilities/JSON.kt
wenxi-zeng: yeah. that's a valid concern. the good thing is that it only overrides set operator for `MutableMap<String, JsonElement>`, not the other types. I tried `MutableMap<String, String>`, it's not impacted. so the impact is limited.

PR #65, opened 2021-12-16
fix useLifecycleObserver not trigger lifecycle events issue

wenxi-zeng:
* fix useLifecycleObserver not trigger lifecycle events issue

### DIFF
--- a/android/src/main/java/com/segment/analytics/kotlin/android/plugins/AndroidLifecyclePlugin.kt
+++ b/android/src/main/java/com/segment/analytics/kotlin/android/plugins/AndroidLifecyclePlugin.kt
@@ -47,16 +47,15 @@ class AndroidLifecyclePlugin() : Application.ActivityLifecycleCallbacks, Default
 
     override fun setup(analytics: Analytics) {
         super.setup(analytics)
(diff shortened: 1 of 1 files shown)


prayansh (approved):

android/src/main/java/com/segment/analytics/kotlin/android/plugins/AndroidLifecyclePlugin.kt
```diff
@@ -47,16 +47,15 @@ class AndroidLifecyclePlugin() : Application.ActivityLifecycleCallbacks, Default
 
     override fun setup(analytics: Analytics) {
         super.setup(analytics)
-        application = analytics.configuration.application as? Application
-            ?: error("no android application context registered")
-        storage = analytics.storage
+        analytics.configuration.let {
```
prayansh, line 50: whats the difference here?

wenxi-zeng: yeah. they're the same. just want to group it together with the other assignments, since they are all visiting properties of `configuration` 😄

PR #67, opened 2022-01-10
prevent using jvm initializer in android and address violation of retrieving unsettable device id

wenxi-zeng:
**Summary**
* prevent using jvm initializer in android
* address violation of retrieving unsettable device id

**Additional Changes**
* add cancel previous to github actions
* updated sovran dependencies from jitpack to maven central
* remove jitpack dependency
* fix android snapshot depending on core release artifacts issue


android/src/main/java/com/segment/analytics/kotlin/android/plugins/AndroidContextPlugin.kt
```diff
@@ -270,4 +242,35 @@ fun hasPermission(context: Context, permission: String): Boolean {
...
+}
+
+
+/**
+ * Workaround for not able to get device id on Android 10 or above using DRM API
+ * {@see https://stackoverflow.com/questions/58103580/android-10-imei-no-longer-available-on-api-29-looking-for-alternatives}
+ * {@see https://developer.android.com/training/articles/user-data-ids}
+ */
+fun getUniqueID(): String? {
+    if (Build.VERSION.SDK_INT < Build.VERSION_CODES.JELLY_BEAN_MR2)
```
prayansh, line 254: given that we support api 16+ we should have some fallback logic here

android/src/test/java/com/segment/analytics/kotlin/android/AndroidAnalyticsKtTest.kt
```diff
@@ -0,0 +1,20 @@
+package com.segment.analytics.kotlin.android
+
+import org.junit.jupiter.api.Assertions.*
+
+import org.junit.jupiter.api.Test
+
+internal class AndroidAnalyticsKtTest {
+    @Test
+    fun `jvm initializer in android platform should failed`() {
+        try {
```
prayansh, line 10: i think the new junit has a way to expect exceptions instead of doing this pattern. if u wanna try that

prayansh (approved):
Looks good. couple comments

android/src/main/java/com/segment/analytics/kotlin/android/plugins/AndroidContextPlugin.kt
wenxi-zeng: good point. the `getDeviceId` method does a null check and returns random uuid as fallback.

android/src/test/java/com/segment/analytics/kotlin/android/AndroidAnalyticsKtTest.kt
wenxi-zeng: let me try that

PR #68, opened 2022-01-12
Fix the enable logic for destinations

prayansh:
* Fix the enable logic for destinations
* Fix SegmentDestination to adhere to `update` API contracts

### DIFF
--- a/core/src/main/java/com/segment/analytics/kotlin/core/platform/plugins/SegmentDestination.kt
+++ b/core/src/main/java/com/segment/analytics/kotlin/core/platform/plugins/SegmentDestination.kt
@@ -101,6 +101,7 @@ class SegmentDestination : DestinationPlugin() {
     }
 
     override fun update(settings: Settings, type: Plugin.UpdateType) {
+        super.update(settings, type)
         if (settings.hasIntegrationSettings(this)) {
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/segment/analytics/kotlin/core/platform/plugins/SegmentDestination.kt
```diff
@@ -101,6 +101,7 @@ class SegmentDestination : DestinationPlugin() {
     }
 
     override fun update(settings: Settings, type: Plugin.UpdateType) {
+        super.update(settings, type)
         if (settings.hasIntegrationSettings(this)) {
```
wenxi-zeng, line 105: since we're calling super, can we reuse `enabled` here?

wenxi-zeng (approved):
LGTM

core/src/main/java/com/segment/analytics/kotlin/core/platform/plugins/SegmentDestination.kt
prayansh: ideally yes, but this is more of a check for the presence of settings, which could be none in the case of manually enabled destinations

PR #70, opened 2022-02-10
Upgraded core build gradle to .kts

Reedyuk:
Minor change, moved build gradle over to a .kts one instead of the groovy one.


wenxi-zeng:
we will have to change all other gradle files to .kts too. however, .kts follows a strict structure in order to recognize other .kts files. we will have to build a plugin to fully transform from groovy. I'd suggest to put this on hold until after GA

Reedyuk:
GA?

wenxi-zeng:
> GA?

general availability. we'll soon be on that phase

prayansh:
@Reedyuk thanks for this upgrade ❤️ . We are currently working on our big release so might be a bit delayed on this change / response.

PR #71, opened 2022-02-15
fix ANR caused by DRM api

wenxi-zeng:
* fix DRM api ANR issue on certain device
* update coroutine to 1.6.0
* make it possible to inject coroutine settings for testing via inheritance

### DIFF
--- a/android/src/main/java/com/segment/analytics/kotlin/android/plugins/AndroidContextPlugin.kt
+++ b/android/src/main/java/com/segment/analytics/kotlin/android/plugins/AndroidContextPlugin.kt
@@ -25,8 +25,10 @@ import java.util.UUID
 import java.lang.System as JavaSystem
 import android.media.MediaDrm
 import com.segment.analytics.kotlin.core.utilities.*
+import kotlinx.coroutines.async
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.withTimeout
+import kotlinx.coroutines.withTimeoutOrNull
 import java.lang.Exception
 import java.security.MessageDigest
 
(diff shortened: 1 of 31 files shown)


prayansh (approved):

android/src/main/java/com/segment/analytics/kotlin/android/plugins/AndroidContextPlugin.kt
```diff
@@ -123,16 +125,43 @@ class AndroidContextPlugin : Plugin {
             emptyJsonObject
         }
 
+        // generate random identifier that does not persist across installations
+        // use it as the default device id in case DRM API failed to generate one.
+        val fallbackDeviceId = UUID.randomUUID().toString()
```
prayansh, line 130: is it better to have no device id or like a string of 0s to indicate not collected. and then use the fallback one when we fail to fetch via DRM?

android/src/main/java/com/segment/analytics/kotlin/android/plugins/AndroidContextPlugin.kt
```diff
@@ -123,16 +125,43 @@ class AndroidContextPlugin : Plugin {
...
+
+        // run `getDeviceId` in coroutine, since the DRM API takes a long time
+        // to generate device id on certain devices and causes ANR issue.
+        analytics.analyticsScope.launch(analytics.analyticsDispatcher) {
+            var deviceId = fallbackDeviceId
+
+            // restrict getDeviceId to 2s to avoid ANR
+            // have to use a different scope than analyticsScope.
+            // otherwise, timeout cancellation won't work (i.e. the scope can't cancel itself)
+            val task = CoroutineScope(SupervisorJob()).async {
```
prayansh, line 147: why do we need to make another coroutineScope here? arent we already in one?

android/src/main/java/com/segment/analytics/kotlin/android/plugins/AndroidContextPlugin.kt
wenxi-zeng: yeah. that's a good call out. my consideration is that in the case when DRM does fail, we have a consistent device id before and after (it takes 2s to know it's failed, could have some cached events). but using all 0s to indicate not ready to collect is also a good idea.

prayansh (approved):
i like the caching ❤️

PR #80, opened 2022-03-24
fix api host being overwritten by default value

wenxi-zeng:
* fix api host being overwritten by default value
* now the apiHost value honors in the following order: remote apiHost > local modified apiHost > default apiHost

### DIFF
--- a/core/src/main/java/com/segment/analytics/kotlin/core/platform/plugins/SegmentDestination.kt
+++ b/core/src/main/java/com/segment/analytics/kotlin/core/platform/plugins/SegmentDestination.kt
@@ -18,7 +18,7 @@ import kotlinx.serialization.json.jsonPrimitive
 @Serializable
 data class SegmentSettings(
     var apiKey: String,
-    var apiHost: String = DEFAULT_API_HOST,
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/segment/analytics/kotlin/core/platform/plugins/SegmentDestination.kt
```diff
@@ -95,8 +95,8 @@ class SegmentDestination : DestinationPlugin(), VersionedPlugin {
     override fun update(settings: Settings, type: Plugin.UpdateType) {
         super.update(settings, type)
         if (settings.hasIntegrationSettings(this)) {
-            settings.destinationSettings<SegmentSettings>(key)?.let {
-                pipeline.apiHost = it.apiHost
+            settings.destinationSettings<SegmentSettings>(key)?.apiHost?.let {
```
prayansh, line 98: OMG thats so subtle of a difference 😓 . Lets add a comment here indicating that we only want to populate the apiHost value if it exists

prayansh (approved):

PR #86, opened 2022-05-11
find destination plugin by name

wenxi-zeng:
* find destination plugin by name


core/src/main/java/com/segment/analytics/kotlin/core/Analytics.kt
```diff
@@ -405,6 +406,12 @@ open class Analytics protected constructor(
      */
     fun <T: Plugin> find(plugin: KClass<T>): T? = this.timeline.find(plugin)
 
+    /**
+     * Retrieve the first match of registered destination plugin by name. It finds
+     * @param destination [String]
+     */
+    fun find(destination: String): DestinationPlugin? = this.timeline.find(destination)
```
prayansh, line 413: let's change param name from destination to `destinationName` / `destinationKey`

prayansh (approved):

PR #87, opened 2022-05-19
address event plugin disparity and track issue

wenxi-zeng:
* address event plugin `execute` disparity between swift and kotlin
* fix issue that event being lost if tracked right after sdk initialization

### DIFF
--- a/core/src/main/java/com/segment/analytics/kotlin/core/platform/Mediator.kt
+++ b/core/src/main/java/com/segment/analytics/kotlin/core/platform/Mediator.kt
@@ -24,13 +24,13 @@ internal class Mediator(internal val plugins: MutableList<Plugin>) {
         var result: BaseEvent? = event
 
         plugins.forEach { plugin ->
-            if (result != null) {
+            result?.let {
                 when (plugin) {
                     is DestinationPlugin -> {
-                        plugin.execute(result!!)
+                        plugin.execute(it)
                     }
                     else -> {
-                        result = plugin.execute(result!!)
+                        result = plugin.execute(it)
                     }
                 }
             }
--- a/core/src/main/java/com/segment/analytics/kotlin/core/platform/Mediator.kt
+++ b/core/src/main/java/com/segment/analytics/kotlin/core/platform/Mediator.kt
@@ -27,29 +27,10 @@ internal class Mediator(internal val plugins: MutableList<Plugin>) {
             if (result != null) {
                 when (plugin) {
                     is DestinationPlugin -> {
-                        plugin.process(result)
-                    }
-                    is EventPlugin -> {
-                        when (result) {
-                            is IdentifyEvent -> {
-                                result = plugin.identify(result as IdentifyEvent)
-                            }
-                            is TrackEvent -> {
-                                result = plugin.track(result as TrackEvent)
-                            }
-                            is GroupEvent -> {
-                                result = plugin.group(result as GroupEvent)
-                            }
-                            is ScreenEvent -> {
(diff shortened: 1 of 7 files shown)


core/src/main/java/com/segment/analytics/kotlin/core/platform/Mediator.kt
```diff
@@ -27,29 +27,10 @@ internal class Mediator(internal val plugins: MutableList<Plugin>) {
...
-                                result = plugin.group(result as GroupEvent)
-                            }
-                            is ScreenEvent -> {
-                                result = plugin.screen(result as ScreenEvent)
-                            }
-                            is AliasEvent -> {
-                                result = plugin.alias(result as AliasEvent)
-                            }
-                        }
+                        plugin.execute(result!!)
```
prayansh, line 30: i think this needs to be `result = plugin.execute(result)`

secondly, we should not be using `!!` since the result from one of the plugins could be null and this would then throw an NPE. lets wrap the call in a null check or use `?.let` and capture the state of result before using it as non-null

core/src/main/java/com/segment/analytics/kotlin/core/platform/Mediator.kt
```diff
@@ -27,29 +27,10 @@ internal class Mediator(internal val plugins: MutableList<Plugin>) {
...
                     }
                     else -> {
-                        result = plugin.execute(result as BaseEvent)
+                        result = plugin.execute(result!!)
```
prayansh, line 33: same as above

core/src/main/java/com/segment/analytics/kotlin/core/platform/Plugin.kt
```diff
@@ -70,6 +70,14 @@ interface EventPlugin : Plugin {
         return payload
     }
 
+    override fun execute(event: BaseEvent): BaseEvent? = when (event) {
```
prayansh, line 73: should this be marked `final`?

prayansh (changes requested):
Left a few comments for changes.

Also are we sure this change wont break any existing plugin implementations? I didnt see any new tests to verify the behavior consistency

core/src/main/java/com/segment/analytics/kotlin/core/platform/Mediator.kt
wenxi-zeng: hrm. `result = plugin.execute(result)` is a compile time error. we don't need a null check because we checked it on line 27

core/src/main/java/com/segment/analytics/kotlin/core/platform/Plugin.kt
wenxi-zeng: no. we shouldn't, because we extends `execute` in edge function. amplitude also overrides this method.

core/src/main/java/com/segment/analytics/kotlin/core/platform/Mediator.kt
prayansh: actually scratch that first comment, didnt realize this was for DestinationPlugin

wenxi-zeng:
I'd say we're pretty safe here. the difference lays on whether the child plugin overrides execute method and if the execute method calls `super.execute`. if not calling `super.execute`, then yes it might break things. so I checked all the existing event plugins. amplitude is the only one that does not call super.execute, but it has a `when` statement in its implementation which basically does the same thing.

prayansh (approved):
🚢

PR #104, opened 2022-08-29
APIs for Loose Map conversion for primitives

prayansh:
Adding support for conversion of most primitive types. Primitive here should mean any types declared in Kotlin SDK or JVM, and not brought in by an external library.

Adds APIs to convert following types to JsonElement equivalent
- String
- Char
- CharArray
- Double
- DoubleArray
- Float
- FloatArray
- Long
- LongArray
- Int
- IntArray
- Short
- ShortArray
- Byte
- ByteArray
- Boolean
- BooleanArray
- Unit
- UInt
- ULong
- UByte
- UShort
- Collection<Any>
- Map<String, Any>
- Array<Any>
- Pair<Any, Any>
- Triple<Any, Any>
- Map.Entry<Any, Any>

Happy to accept future contributions to increase this type coverage

### DIFF
--- a/core/src/main/java/com/segment/analytics/kotlin/core/utilities/JSON.kt
+++ b/core/src/main/java/com/segment/analytics/kotlin/core/utilities/JSON.kt
@@ -2,9 +2,21 @@
 
 package com.segment.analytics.kotlin.core.utilities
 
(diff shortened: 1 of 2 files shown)


wenxi-zeng (approved):
🚀

PR #107, opened 2022-08-31
add api for identify using only traits

prayansh:
Adds new api `fun Analytics.identify(traits)` that allows users to update traits without needing userId


wenxi-zeng (approved):
LGTM. can we also have these apis in java compat?

PR #109, opened 2022-09-12
Correct minsdkversion

niallzato:
Our timestamps [rely](https://github.com/segmentio/analytics-kotlin/blob/ed398f090cafb98d42a93ff4c5c4cff4b1323885/core/src/main/java/com/segment/analytics/kotlin/core/Events.kt#L13) on this class: https://developer.android.com/reference/java/time/Instant.html

Which requires API level 26 or above, changing our minSdkVersion to reflect this.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -13,7 +13,7 @@ android {
         // Required when setting minSdkVersion to 20 or lower
         multiDexEnabled true
 
-        minSdkVersion 16
+        minSdkVersion 26
         targetSdkVersion 31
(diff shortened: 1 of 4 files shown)


prayansh:
Hey @niallzato! We are using the `coreLibraryDesugaring 'com.android.tools:desugar_jdk_libs:1.1.5'` gradle config (More info here: https://developer.android.com/studio/write/java8-support#library-desugaring), enabling us to use `Instant` class in API levels < 26.

PR #114, opened 2022-10-03
Fix `_metadata` population

prayansh:
Our previous implementation was lacking a little, causing issues with sending data to destinations on cloud-mode.

These changes should now incorporate the list of active destinations and mark them as `unbundled` if no device-mode equivalent is present.


wenxi-zeng (approved):
LGTM. couple minor optimizations:
1. make `bundled` as a set first. so `bundled.contains` becomes an O(1) operation, and filter + contains is O(n). otherwise it is O(mn).
2. make `unbundled` as a set first. so we don't have to do a union at the end, saves m+n operations, otherwise, it's 2*(m+n).

PR #124, opened 2022-12-07
LIBMOBILE-1060

charleswilliamson7:


core/src/main/java/com/segment/analytics/kotlin/core/platform/plugins/UserInfoPlugin.kt
```diff
@@ -14,27 +13,32 @@ import kotlinx.serialization.json.put
 class UserInfoPlugin : Plugin {
     override val type: Plugin.Type = Plugin.Type.Before
     override lateinit var analytics: Analytics
+    private lateinit var newUser: JsonObject
```
wenxi-zeng, line 16: `newUser` is declared but never assigned

core/src/main/java/com/segment/analytics/kotlin/core/platform/plugins/UserInfoPlugin.kt
```diff
@@ -14,27 +13,32 @@ import kotlinx.serialization.json.put
...
-    private fun applyUserInfoData(event: BaseEvent) {
-        val newUserInfo = buildJsonObject {
-            // copy existing context
-            put("anonymousId", event.anonymousId)
-            put("userId", event.userId)
-        }
+    override fun execute(event: BaseEvent): BaseEvent {
+        store = analytics.store
+        userId = newUser.get("userId").toString()
+        anonymousId = newUser.get("anonymousId").toString()
```
wenxi-zeng, line 25: reading a variable that is never initialized and assigned

wenxi-zeng (changes requested):

core/src/main/java/com/segment/analytics/kotlin/core/platform/plugins/UserInfoPlugin.kt
```diff
@@ -0,0 +1,39 @@
...
+            analytics.userInfo.userId =  injectedEvent.userId
+            analytics.userInfo.anonymousId =  injectedEvent.anonymousId
+
+        } else if (injectedEvent.type === EventType.Alias) {
+
+            analytics.userInfo.anonymousId =  injectedEvent.anonymousId
+
+        } else {
+
+            injectedEvent.userId = analytics.userInfo.userId.toString()
```
wenxi-zeng, line 31: shouldn't use toString here. it will cause userid to be string value 'null'. we need to make it empty string if it is null. see [here](https://github.com/segmentio/analytics-kotlin/blob/main/core/src/main/java/com/segment/analytics/kotlin/core/Events.kt#L113) for example

wenxi-zeng (commented):
the following changes should be made in Analytics:
1. update `userId()` to return userId directly from the cache
2. update `userIdAsync` to return userId directly from the cache, and mark this function as deprecate, suggest the user to use `userId()` instead
3. repeat the 1-2 above for `anonymousId`

core/src/main/java/com/segment/analytics/kotlin/core/Analytics.kt
```diff
@@ -536,39 +546,41 @@ open class Analytics protected constructor(
     }
 
     /**
-     * Retrieve the userId registered by a previous `identify` call in a blocking way.
-     * Note: this method invokes `runBlocking` internal, it's not recommended to be used
-     * in coroutines.
+     * Retrieve the userId registered by a previous `identify` call.
      */
     @BlockingApi
```
wenxi-zeng, line 551: this annotation should be removed

core/src/main/java/com/segment/analytics/kotlin/core/Analytics.kt
```diff
@@ -536,39 +546,41 @@ open class Analytics protected constructor(
...
+        return userId()
     }
 
     /**
-     * Retrieve the traits registered by a previous `identify` call in a blocking way.
-     * Note: this method invokes `runBlocking` internal, it's not recommended to be used
-     * in coroutines.
+     * Retrieve the traits registered by a previous `identify` call.
      */
     @BlockingApi
```
wenxi-zeng, line 570: this annotation should be removed

core/src/main/java/com/segment/analytics/kotlin/core/Analytics.kt
```diff
@@ -536,39 +546,41 @@ open class Analytics protected constructor(
...
-     * Retrieve the userId registered by a previous `identify` call in a blocking way.
-     * Note: this method invokes `runBlocking` internal, it's not recommended to be used
-     * in coroutines.
+     * Retrieve the userId registered by a previous `identify` call.
      */
     @BlockingApi
-    fun userId(): String? = runBlocking {
-        userIdAsync()
+    fun userId(): String? {
+        return userInfo.userId
```
wenxi-zeng, line 553: need to do the same for `anonymousId`

core/src/main/java/com/segment/analytics/kotlin/core/Analytics.kt
```diff
@@ -536,39 +546,41 @@ open class Analytics protected constructor(
...
      * Retrieve the userId registered by a previous `identify` call
      */
+    @Deprecated(
+        "This function no longer serves a purpose and internally calls `userId()`.",
+        ReplaceWith("userId()")
+    )
     suspend fun userIdAsync(): String? {
-        val userInfo = store.currentState(UserInfo::class)
-        return userInfo?.userId
+        return userId()
```
wenxi-zeng, line 564: this call no longer need to be `suspend`, same as the other ones

core/src/main/java/com/segment/analytics/kotlin/core/Analytics.kt
```diff
@@ -536,39 +546,41 @@ open class Analytics protected constructor(
...
      * Retrieve the traits registered by a previous `identify` call
      */
+    @Deprecated(
+        "This function no longer serves a purpose and internally calls `traits()`.",
+        ReplaceWith("traits()")
+    )
     suspend fun traitsAsync(): JsonObject? {
-        val userInfo = store.currentState(UserInfo::class)
-        return userInfo?.traits
+        return traits()
```
wenxi-zeng, line 583: need to update the other two `traits` and `traitsAsync` calls below.

core/src/main/java/com/segment/analytics/kotlin/core/Analytics.kt
```diff
@@ -527,6 +533,10 @@ open class Analytics protected constructor(
      * user logs out
      */
     fun reset() {
+        userInfo.userId = null
```
wenxi-zeng, line 536: wonder if we could just new a `UserInfo` object instead of doing assignments individually?

wenxi-zeng (approved):

wenxi-zeng (approved):
good catch!

PR #126, opened 2022-12-13
Adding a traits injector plugin sample.

didiergarcia:


samples/kotlin-android-app/src/main/java/com/segment/analytics/next/plugins/InjectTraitsPlugin.kt
```diff
@@ -0,0 +1,22 @@
...
+    override val type: Plugin.Type = Plugin.Type.Enrichment
+    override lateinit var analytics: Analytics
+
+    override fun execute(event: BaseEvent): BaseEvent? {
+
+        if (event.type == EventType.Identify) {
+            return event
+        }
+
+        event.context.plus(Pair("traits", analytics.traits()));
```
wenxi-zeng, line 18: I believe `plus` is an immutable call. try the following
```
        event.context = updateJsonObject(event.context) {
            it["traits"] = analytics.traits()!!
        }
```

also, `analytics.traits()` is a blocking call, we should avoid using it. one way to do it is to keep a traits copy in the plugin, and update it for every Identify call. see [this](https://github.com/segmentio/analytics-react-native/pull/660/files#diff-ceb59bec11b02497336b7b3a22298e56d35a7b981b0e0cc7d2ad24d2a79f34db) for an example

PR #133, opened 2023-02-13
Feature/add full deep link support to kotlin sample

didiergarcia:
Add full support for deep links in Kotlin Analytics. 

This adds a new top-level API `addDeepLinkOpen()` function to the Analytics functions that people will use to manually track the app opening from a deep link.

### DIFF
--- a/samples/kotlin-android-app/src/main/java/com/segment/analytics/next/MainActivity.kt
+++ b/samples/kotlin-android-app/src/main/java/com/segment/analytics/next/MainActivity.kt
@@ -7,7 +7,7 @@ import androidx.appcompat.app.AppCompatActivity
 import androidx.fragment.app.Fragment
 import androidx.fragment.app.FragmentManager
 import androidx.fragment.app.FragmentTransaction
-import com.segment.analytics.kotlin.android.addDeepLinkOpen
+import com.segment.analytics.kotlin.android.trackDeepLinkOpen
 import com.segment.analytics.kotlin.core.Analytics
 import com.segment.analytics.kotlin.core.EventType
 import com.segment.analytics.kotlin.core.platform.plugins.logger.log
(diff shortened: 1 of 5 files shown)


samples/kotlin-android-app/src/main/java/com/segment/analytics/next/MainActivity.kt
```diff
@@ -52,4 +54,18 @@ class MainActivity : AppCompatActivity() {
...
+
+        // Add a manual deep-link opened event.
+        // This is necessary when your Activity has a android:launchMode of
+        // 'singleInstance', 'singleInstancePerTask', 'singleTop', or any other mode
+        // that will re-use an existing Activity instead of creating a new instance.
+        // The Analytics SDK automatically identifies when you app is started from a deep link
+        // if the Activity is created, but not if it is re-used. Therefore we have to add this
+        // code to manually capture the Deep Link info.
+
+        Analytics.addDeepLinkOpen(analytics, "deep-link", intent)
```
wenxi-zeng, line 69: out of curiosity, why do we choose to add an extend function for `Analytics.Companion` instead for `Analytics`? if we extend `Analytics`, we can do something like
```
analytics.addDeepLinkOpen("deep-link", intent)
```
it seems cleaner 😄 
also, would `trackDeepLinkOpen` be a better name? feels more consistent than the `track` calls

wenxi-zeng (approved):
LGTM

samples/kotlin-android-app/src/main/java/com/segment/analytics/next/MainActivity.kt
didiergarcia: Great Questions!

Why add the function on `Analytics` (static) instead `analytics`:

So this was a conscious choice because in looking at code in `MainActivity`it had both types of calls:

```
Analytics.debugLogsEnabled = true
```

And 

```
analytics.flush()
```

I decided to go with static function because I thought it made sense, but it's not a hard decision to change. I will update it!

Regarding the name of the function: I like your name better. Consider it changed.

wenxi-zeng (approved):

PR #146, opened 2023-03-10
Add CoroutineExceptionHandler and try/catch blocks over critical area…

didiergarcia:
…s. Also moved AndroidLogger setup to earlier time to handle early crashes.


core/src/main/java/com/segment/analytics/kotlin/core/Analytics.kt
```diff
@@ -99,27 +102,31 @@ open class Analytics protected constructor(
...
+                    add(SegmentDestination())
+                }
 
-            if (configuration.autoAddSegmentDestination) {
-                add(SegmentDestination())
+                checkSettings()
             }
-
-            checkSettings()
+        } catch (t: Throwable) {
```
wenxi-zeng, line 128: maybe we don't have to include analyticsScope.launch in the try block, since the scope has its own handler?

wenxi-zeng (approved):

core/src/main/java/com/segment/analytics/kotlin/core/Analytics.kt
didiergarcia: Good callout. Moved this into the timeline.add() function which is what ends up calling the plugin.setup().

PR #154, opened 2023-04-19
Remove Desugaring Requirement By Removing Instant.now()

didiergarcia:


core/src/main/java/com/segment/analytics/kotlin/core/Analytics.kt
```diff
@@ -89,6 +89,7 @@ open class Analytics protected constructor(
                 Analytics.segmentLog(
                     "Caught Exception in Analytics Scope: ${t}"
                 )
+                t.printStackTrace()
```
wenxi-zeng, line 92: do we need this here?

android/src/test/java/com/segment/analytics/kotlin/android/EventsFileTests.kt
```diff
@@ -32,6 +33,8 @@ class EventsFileTests {
 
     init {
         mockkStatic(Instant::class)
+        mockkStatic(::dateTimeNowString)
+        every { dateTimeNowString() } returns Date(0).toInstant().toString()
```
wenxi-zeng, line 37: do we need to replace `toInstant` too? looks like it's returning an `Instant` value. would that require desugaring?

wenxi-zeng (approved):
LGTM. let's remove all the `desugaring` setup and dependencies before merge in

android/src/test/java/com/segment/analytics/kotlin/android/EventsFileTests.kt
didiergarcia: Not in our tests. I let all of those keep the use of the Instant classes. Instant was only removed from code that runs on the device.

didiergarcia: Yes. It is still available because the test run locally using java 11. I have updated the code to remove desugaring from the android build.gradle and re-run the test and they still all pass.

PR #158, opened 2023-04-28
LIBMOBILE-1277- fix androidx lifecycle causes analytics init in background crash

RawatRaveena:
LIBMOBILE-1277
- fix androidx lifecycle causes analytics init in background crash

### DIFF
--- a/android/src/main/java/com/segment/analytics/kotlin/android/plugins/AndroidLifecyclePlugin.kt
+++ b/android/src/main/java/com/segment/analytics/kotlin/android/plugins/AndroidLifecyclePlugin.kt
@@ -7,8 +7,7 @@ import android.content.pm.PackageInfo
 import android.content.pm.PackageManager
 import android.net.ParseException
 import android.net.Uri
-import android.os.Build
-import android.os.Bundle
+import android.os.*
 import androidx.lifecycle.DefaultLifecycleObserver
 import androidx.lifecycle.Lifecycle
 import androidx.lifecycle.LifecycleObserver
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/segment/analytics/kotlin/android/plugins/AndroidLifecyclePlugin.kt
```diff
@@ -260,14 +264,25 @@ class AndroidLifecyclePlugin() : Application.ActivityLifecycleCallbacks, Default
         application.unregisterActivityLifecycleCallbacks(this)
         if (useLifecycleObserver) {
             // only unregister if feature is enabled
-            lifecycle.removeObserver(this)
+            run(Runnable { lifecycle.removeObserver(this) })
+        }
+    }
+    private fun run(runnable: Runnable) {
```
wenxi-zeng, line 270: you don't have to use runnable in kotlin, try [high order function and lambda](https://kotlinlang.org/docs/lambdas.html)

android/src/main/java/com/segment/analytics/kotlin/android/plugins/AndroidLifecyclePlugin.kt
```diff
@@ -260,14 +264,25 @@ class AndroidLifecyclePlugin() : Application.ActivityLifecycleCallbacks, Default
...
+        if (Looper.myLooper() == Looper.getMainLooper()) {
+            runnable.run()
+        } else {
+            HANDLER.post(runnable)
```
wenxi-zeng, line 274: in kotlin, the correct way to post to main thread is 
```
launch(Dispatchers.Main) {
  // do something on main thread
}
```

wenxi-zeng (changes requested):

didiergarcia (approved):

wenxi-zeng (approved):

PR #168, opened 2023-06-01
Support Map<String, Any> for properties and traits.

didiergarcia:


core/build.gradle
```diff
@@ -24,13 +24,13 @@ test {
 dependencies {
     // MAIN DEPS
     api 'com.segment:sovran-kotlin:1.2.2'
-    api "org.jetbrains.kotlinx:kotlinx-serialization-json:1.4.1"
-    implementation 'org.jetbrains.kotlinx:kotlinx-coroutines-core:1.6.4'
+    api "org.jetbrains.kotlinx:kotlinx-serialization-json:1.5.1"
```
wenxi-zeng, line 27: update the kotlin libs version. need to do the same for android gradle

core/src/main/java/com/segment/analytics/kotlin/core/Analytics.kt
```diff
@@ -170,15 +171,14 @@ open class Analytics protected constructor(
      * @param properties to describe the action. Needs to be [serializable](https://github.com/Kotlin/kotlinx.serialization/blob/master/docs/serializers.md)
      * @see <a href="https://segment.com/docs/spec/track/">Track Documentation</a>
      */
-    inline fun <reified T : Any> track(
+    inline fun <reified T> track(
```
wenxi-zeng, line 174: removed `Any`. this was what causes Any serializer not found

core/src/main/java/com/segment/analytics/kotlin/core/Analytics.kt
```diff
@@ -170,15 +171,14 @@ open class Analytics protected constructor(
...
         name: String,
         properties: T,
     ) {
-        if (properties is Map<*, *>) {
-            track(name, properties, AnySerializer)
-        } else {
-            track(name, properties, Json.serializersModule.serializer())
-        }
+        val format = Json { serializersModule = SerializersModule {
```
wenxi-zeng, line 178: need to move `format` to utilities. we gonna need it for the other calls

core/src/main/java/com/segment/analytics/kotlin/core/utilities/AnySerializer.kt
```diff
@@ -16,8 +15,9 @@ object AnySerializer: KSerializer<Any> {
         return "not-implemented";
     }
 
-    @OptIn(InternalSerializationApi::class)
-    override val descriptor: SerialDescriptor = buildSerialDescriptor("My Any Serializer", SerialKind.CONTEXTUAL)
+    @OptIn(InternalSerializationApi::class, ExperimentalSerializationApi::class)
+    override val descriptor: SerialDescriptor = ContextualSerializer(Any::class, null, emptyArray()).descriptor
```
wenxi-zeng, line 19: I think this change doesn't matter. just nice to have, instead of init it arbitrary

core/src/main/java/com/segment/analytics/kotlin/core/utilities/AnySerializer.kt
```diff
@@ -0,0 +1,34 @@
...
+    override fun serialize(encoder: Encoder, value: Any) {
+        val toJsonElement = value.toJsonElement()
+        encoder.encodeSerializableValue(Json.serializersModule.serializer(), toJsonElement)
+    }
+}
+
+/**
+ * A pre-configured Json Implementation with an Any type serializer.
+ */
+val JsonSerializer = Json { serializersModule = SerializersModule {
```
wenxi-zeng, line 32: `JsonSerializer` feels a bit too more generic. I was confused it with `Json` when I read the code. feel `AnySerializer` is a better name, even we kind randomly named it. Can we change it to something else that reflects its actual use case?

wenxi-zeng (approved):

core/src/main/java/com/segment/analytics/kotlin/core/utilities/AnySerializer.kt
didiergarcia: Ugh... name conflicts with the `AnySerializer` that does the actually work. I changed it to `JsonAnySerializer`.

PR #170, opened 2023-06-05
Locale Timestamp

dennisdeng2002:
On Android, if you set the app language to `Arabic` the timestamp will be formatted using the device locale, i.e. `٠٥T١٠:٤٨:٢٩.٥Z`. I've attached screenshots of the behavior before / after the changes

**Before / After (Arabic)**

 

**Before / After (English)**

### DIFF
--- a/core/src/main/java/com/segment/analytics/kotlin/core/utilities/DateTimeUtils.kt
+++ b/core/src/main/java/com/segment/analytics/kotlin/core/utilities/DateTimeUtils.kt
@@ -12,7 +12,7 @@ import java.util.*
  * dateTimeNowString(): 2023-04-19T04:03:46.880Z
  */
 fun dateTimeNowString(): String {
-    val sdf = SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss'.'Szzz")
+    val sdf = SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss'.'Szzz", Locale.ROOT)
     val utc = TimeZone.getTimeZone("UTC");
     sdf.timeZone = utc;
     return sdf.format(Date()).replace("UTC", "Z")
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/segment/analytics/kotlin/core/utilities/DateTimeUtils.kt
```diff
@@ -12,8 +12,10 @@ import java.util.*
  * dateTimeNowString(): 2023-04-19T04:03:46.880Z
  */
 fun dateTimeNowString(): String {
-    val sdf = SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss'.'Szzz")
+    // Note, we should specify locale = Locale.ROOT, otherwise the timestamp returned will use
+    // the default locale, which may not be what we want.
+    val sdf = SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss'.'Szzz", Locale.ROOT)
```
dennisdeng2002, line 17: The [Android docs](https://developer.android.com/reference/java/util/Locale) recommend using `Locale.US`, but since this might be used in JVM projects I opted to go with `Locale.ROOT`

> The default locale is not appropriate for machine-readable output. The best choice there is usually Locale.US – this locale is guaranteed to be available on all devices, and the fact that it has no surprising special cases and is frequently used (especially for computer-computer communication) means that it tends to be the most efficient choice too.

wenxi-zeng (approved):

core/src/test/kotlin/com/segment/analytics/kotlin/core/utilities/DateTimeUtilsTest.kt
```diff
@@ -0,0 +1,14 @@
...
+
+import org.junit.jupiter.api.Assertions.*
+import org.junit.jupiter.api.Test
+
+class DateTimeUtilsTest {
+
+    @Test
+    fun `dateTimeNowString() produces a string in the correct format`() {
+        val dateTimeNowString = dateTimeNowString()
+        val regex = Regex("^[0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}:\\d{2}\\.\\d{3}Z$")
```
wenxi-zeng, line 11: this regex fails unit test. try this one `"^\\d{4}-[0-1][0-3]-[0-3]\\d{1}T[0-2]\\d{1}:[0-5]\\d{1}:[0-5]\\d{1}Z$"`

PR #172, opened 2023-06-15
Custom Android Logger

dennisdeng2002:
Add the ability to set a custom `AndroidLogger`. Technically people can directly call `Analytics.setLogger()` but I think it's cleaner to add it explicitly to the API


dennisdeng2002:
cc @wenxi-zeng @didiergarcia

wenxi-zeng:
hey @dennisdeng2002, thanks for the pr. but I don't think this change is necessary cleaner, instead it leads to confusion and bugs, since `logger` is a singleton object, but this pr proposes to update it at a per instance level. in the case where there are multiple analytics instances, the last created analytics object overwrites the logger set by the other analytics instances. this is a side effect and can be undesirable. thus, `Analytics.setLogger` must be called explicitly

PR #177, opened 2023-07-05
allow custom error handler

wenxi-zeng:
* add custom error handler
* add 3 new APIs
  * `reportInternalError` to direct exceptions to user provided error handler
  * `pendingUploads` to get the yet uploaded files
  * `purgeStorage` to remove all local files
  * `purgeStorage(filePath: String)` to remove a specific file

### DIFF
--- a/android/src/main/java/com/segment/analytics/kotlin/android/plugins/AndroidLifecyclePlugin.kt
+++ b/android/src/main/java/com/segment/analytics/kotlin/android/plugins/AndroidLifecyclePlugin.kt
@@ -64,8 +64,9 @@ class AndroidLifecyclePlugin() : Application.ActivityLifecycleCallbacks, Default
         packageInfo = try {
             packageManager.getPackageInfo(application.packageName, 0)
         } catch (e: PackageManager.NameNotFoundException) {
-            analytics.reportInternalError(e)
-            throw AssertionError("Package not found: " + application.packageName)
+            val error = AssertionError("Package not found: " + application.packageName)
+            analytics.reportInternalError(error)
+            throw error
         }
 
         // setup lifecycle listeners
--- a/android/src/main/java/com/segment/analytics/kotlin/android/plugins/AndroidLifecyclePlugin.kt
+++ b/android/src/main/java/com/segment/analytics/kotlin/android/plugins/AndroidLifecyclePlugin.kt
@@ -18,6 +18,7 @@ import com.segment.analytics.kotlin.android.utilities.DeepLinkUtils
 import com.segment.analytics.kotlin.core.Analytics
 import com.segment.analytics.kotlin.core.Storage
 import com.segment.analytics.kotlin.core.platform.Plugin
+import com.segment.analytics.kotlin.core.reportInternalError
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.launch
 import kotlinx.serialization.json.buildJsonObject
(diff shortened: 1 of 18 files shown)


core/src/main/java/com/segment/analytics/kotlin/core/platform/policies/FrequencyFlushPolicy.kt
```diff
@@ -6,7 +6,7 @@ import kotlinx.coroutines.delay
 import kotlinx.coroutines.isActive
 import kotlinx.coroutines.launch
 
-class FrequencyFlushPolicy(val flushIntervalInMillis: Long = 30_000): FlushPolicy {
+class FrequencyFlushPolicy(var flushIntervalInMillis: Long = 30_000): FlushPolicy {
```
didiergarcia, line 9: Did this have to change to a `var`? I don't see it used in the PR.

core/src/main/java/com/segment/analytics/kotlin/core/State.kt
```diff
@@ -31,7 +31,7 @@ data class System(
                     Settings.serializer(),
                     storage.read(Storage.Constants.Settings) ?: ""
                 )
-            } catch (ex: Exception) {
+            } catch (ignored: Exception) {
```
didiergarcia, line 34: How do we decide which errors we report and which we ignore?

android/src/main/java/com/segment/analytics/kotlin/android/plugins/AndroidLifecyclePlugin.kt
```diff
@@ -63,6 +64,7 @@ class AndroidLifecyclePlugin() : Application.ActivityLifecycleCallbacks, Default
         packageInfo = try {
             packageManager.getPackageInfo(application.packageName, 0)
         } catch (e: PackageManager.NameNotFoundException) {
+            analytics.reportInternalError(e)
```
didiergarcia, line 67: Should we report the `AssertionError` instead of the incoming `NameNotFoundException`?

core/src/main/java/com/segment/analytics/kotlin/core/Analytics.kt
```diff
@@ -670,6 +670,41 @@ open class Analytics protected constructor(
...
+    /**
+     * Purge all pending event upload files.
+     */
+    fun purgeStorage() {
+        // do not call purgeStorage(filePath: String)
+        // since we want to schedule all files deletion in the same dispatch all at once
+        analyticsScope.launch(fileIODispatcher) {
+            for (file in pendingUploads()) {
+                try {
+                    storage.removeFile(file)
```
didiergarcia, line 690: Is there maybe a way to remove the folder of pending upload files, instead of each file?

core/src/main/java/com/segment/analytics/kotlin/core/Errors.kt
```diff
@@ -0,0 +1,21 @@
...
+
+fun Analytics.Companion.reportInternalError(error: Throwable) {
+    error.message?.let {
+        Analytics.segmentLog(it)
+    }
+}
+
+fun Analytics.Companion.reportInternalError(error: String) {
+    Analytics.segmentLog(error)
+}
```
didiergarcia, line 21: Do we ever not want to send the error to the `errorHandler`?

didiergarcia (approved):

core/src/main/java/com/segment/analytics/kotlin/core/Analytics.kt
wenxi-zeng: java.io might have an api to delete a collection. but for us to implement it, we will have to add api to `EventFileManager`, `Storage` and all the implementations of `Storage`. not worth the effort, since the java.oi api eventually delete files one by one

core/src/main/java/com/segment/analytics/kotlin/core/State.kt
wenxi-zeng: for the ones expect a result we should ignore the exception. for example, parsing int from string with default, if an parse exception raises, we should give the default value, instead of raising exception

PR #214, opened 2024-03-22
Telemetry Metrics

MichaelGHSeg:


core/src/main/java/com/segment/analytics/kotlin/core/Telemetry.kt
```diff
@@ -0,0 +1,272 @@
...
+        }
+
+    private val queue = mutableListOf<RemoteMetric>()
+    private var queueBytes = 0
+    private var queueSizeExceeded = false
+    private val seenErrors = mutableMapOf<String, Int>()
+    private var started = false
+
+    init {
+        CoroutineScope(Dispatchers.Default).launch {
```
wenxi-zeng, line 114: we should use our own scope and dispatcher to avoid leaking exceptions to app's scope.

core/src/main/java/com/segment/analytics/kotlin/core/Telemetry.kt
```diff
@@ -0,0 +1,272 @@
...
+                } catch (e: CancellationException) {
+                    // last flush before shutdown
+                    flush()
+                }
+
+            }
+        }
+    }
+
+    fun start() {
```
wenxi-zeng, line 133: it's probably better to launch the flush task in this function?

core/src/main/java/com/segment/analytics/kotlin/core/platform/Mediator.kt
```diff
@@ -44,6 +47,11 @@ internal class Mediator(internal var plugins: CopyOnWriteArrayList<Plugin> = Cop
                 } catch (t: Throwable) {
                     Analytics.reportInternalError("Caught Exception in plugin: $t")
                     Analytics.segmentLog("Skipping plugin due to Exception: $plugin", kind = LogKind.WARNING)
+                    Telemetry.error("analytics_mobile.integration.invoke.error",
```
wenxi-zeng, line 50: can we combine all these three logging mechanism to be one call? feels redundant to see all 3 logging for the same error everywhere

core/src/main/java/com/segment/analytics/kotlin/core/platform/Mediator.kt
```diff
@@ -58,6 +66,11 @@ internal class Mediator(internal var plugins: CopyOnWriteArrayList<Plugin> = Cop
             } catch (t: Throwable) {
                 Analytics.reportInternalError(t)
                 Analytics.segmentLog("Caught Exception applying closure to plugin: $it: $t")
+                Telemetry.error("analytics_mobile.integration.invoke.error",
+                    mapOf("error" to t.toString(), "plugin" to "${it.type}-${it.javaClass}",
```
wenxi-zeng, line 70: better to make `analytics_mobile.integration.invoke.error` as a constant

core/src/main/java/com/segment/analytics/kotlin/core/Telemetry.kt
```diff
@@ -0,0 +1,272 @@
...
+    private var MAX_QUEUE_BYTES = 28000
+    private var _maxQueueBytes: Int = MAX_QUEUE_BYTES
+    private var _httpClient: HTTPClient = HTTPClient("", MetricsRequestFactory())
+    private var _sendWriteKeyOnError: Boolean = true
+    private var _sendErrorLogData: Boolean = false
+    private var _errorHandler: ((Throwable) -> Unit)? = null
+    private var _rateLimitTimer: Int = 0
+
+    var enable: Boolean
+        get() = _enable
```
wenxi-zeng, line 50: no need to write getter/setter out, by default it's property, the compiler generates getter/setter and field

core/src/main/java/com/segment/analytics/kotlin/core/Telemetry.kt
```diff
@@ -0,0 +1,272 @@
...
+        } else {
+            queue.forEach {
+                it.value = (it.value / sampleRate).roundToInt()
+            }
+        }
+
+        started = true
+    }
+
+    fun increment(metric: String, tags: Map<String, String>) {
```
wenxi-zeng, line 147: probably should use enum for metric types?

core/src/main/java/com/segment/analytics/kotlin/core/Settings.kt
```diff
@@ -118,5 +128,9 @@ internal fun Analytics.fetchSettings(
         "${ex.message}: failed to fetch settings",
         kind = LogKind.ERROR
     )
+    Telemetry.error("analytics_mobile.invoke.error",
```
didiergarcia, line 131: Should those Strings be CONST somewhere? Maybe we could add more context with a comment about them:

```
// This is for XYZ...
const INVOKE_ERROR = "analytics_mobile.invoke.error"
````

File: core/src/main/java/com/segment/analytics/kotlin/core/Telemetry.kt [hunk 8]
```diff
@@ -0,0 +1,272 @@
...
+)
+
+fun logError(err: Throwable) {
+    println("Error sending segment performance metrics $err")
+}
+
+object Telemetry {
+    private var _enable: Boolean = true
+    private var _host: String = Constants.DEFAULT_API_HOST
+    private var _sampleRate: Double = 1.0
```
didiergarcia, line 38: that's 100%? Maybe needs a comment to spell that out?

core/src/main/java/com/segment/analytics/kotlin/core/Telemetry.kt
didiergarcia: Do you have to remove the "_" to indicate a backing property and make it non private?

I guess if the only thing we're doing is passing the value without modification it makes sense to make a regular property.

core/src/main/java/com/segment/analytics/kotlin/core/Telemetry.kt
```diff
@@ -0,0 +1,272 @@
...
+    var maxQueueSize: Int
+        get() = _maxQueueSize
+        set(value) {
+            _maxQueueSize = value
+        }
+
+    var maxQueueBytes: Int
+        get() = _maxQueueBytes
+        set(value) {
+            _maxQueueBytes = min(value, MAX_QUEUE_BYTES)
```
didiergarcia, line 98: Oh yeah, Here's one that should have it's `set` defined.

core/src/main/java/com/segment/analytics/kotlin/core/Telemetry.kt
```diff
@@ -0,0 +1,272 @@
...
+                }
+
+            }
+        }
+    }
+
+    fun start() {
+        if (started) return
+        // Assume sampleRate is now set and everything in the queue hasn't had it applied
+        if (Math.random() > sampleRate) {
```
didiergarcia, line 136: `>=` no ?

core/src/main/java/com/segment/analytics/kotlin/core/Telemetry.kt
didiergarcia: The down-side to enum is that you can't add a new one if we're running code from the QuickJS Engine. I don't know if we ever will, but maybe we should use constants for known ones?

core/src/main/java/com/segment/analytics/kotlin/core/Telemetry.kt
```diff
@@ -0,0 +1,272 @@
...
+        }
+
+        started = true
+    }
+
+    fun increment(metric: String, tags: Map<String, String>) {
+        if (!enable) return
+        if (!metric.startsWith("analytics_mobile.")) return
+        if (tags.isEmpty()) return
+        if (Math.random() > sampleRate) return
```
didiergarcia, line 151: `>=` ?

core/src/main/java/com/segment/analytics/kotlin/core/Telemetry.kt
```diff
@@ -0,0 +1,272 @@
...
+    }
+
+    fun error(metric:String, tags: Map<String, String>, log: String) {
+        if (!enable) return
+        if (!metric.startsWith("analytics_mobile.")) return
+        if (tags.isEmpty()) return
+        if (queue.size >= maxQueueSize) return
+
+        var filteredTags = tags
+        if (!sendWriteKeyOnError) filteredTags = tags.filterKeys { it != "writekey" }
```
didiergarcia, line 164: We use about the case here? do we need to toLowerCase() the key before checking?

core/src/main/java/com/segment/analytics/kotlin/core/Telemetry.kt
```diff
@@ -0,0 +1,272 @@
...
+                flush()
+                seenErrors[errorKey] = 0 // Zero because it's already been sent.
+            }
+        }
+    }
+
+    fun flush() {
+        if (!started || !enable || queue.isEmpty()) return
+
+        if (_rateLimitTimer > 0 && _rateLimitTimer > (System.currentTimeMillis() / 1000).toInt()) {
```
didiergarcia, line 187: I think the first check (`> 0`) is redundant?

core/src/main/java/com/segment/analytics/kotlin/core/Telemetry.kt
```diff
@@ -0,0 +1,272 @@
...
+            val connection = httpClient.upload(host)
+            connection.outputStream?.use { outputStream ->
+                // Write the JSON string to the outputStream.
+                outputStream.write(payload.toByteArray(Charsets.UTF_8))
+                outputStream.flush() // Ensure all data is written
+            }
+            connection.close()
+        } catch (e: HTTPException) {
+            _errorHandler?.let { it(e) }
+            if (e.responseCode != 429) {
```
didiergarcia, line 215: Do do we do in the case of Rate Limiting?

core/src/main/java/com/segment/analytics/kotlin/core/Telemetry.kt
```diff
@@ -0,0 +1,272 @@
...
+                val headers = e.responseHeaders
+                val rateLimit = headers["X-RateLimit-Reset"]?.firstOrNull()?.toLongOrNull()
+                if (rateLimit != null) {
+                    _rateLimitTimer = rateLimit.toInt() + (System.currentTimeMillis() / 1000).toInt()
```
didiergarcia, line 219: does that `toInt()` cause issues when it was a `long` coming in?

core/src/main/java/com/segment/analytics/kotlin/core/Telemetry.kt
```diff
@@ -0,0 +1,272 @@
...
+                }
+            }
+        } catch (e: Exception) {
+            _errorHandler?.let { it(e) }
+        }
+    }
+
+    private val additionalTags: Map<String, String> by lazy {
+        var osVersion = System.getProperty("os.version")
```
didiergarcia, line 228: There is probably already code to get this information somewhere?

core/src/main/java/com/segment/analytics/kotlin/core/Telemetry.kt
```diff
@@ -0,0 +1,272 @@
...
+        val found = queue.find {
+            it.metric == metric && it.tags == (tags + additionalTags)
+        }
+        if (found != null) {
+            found.value += value
+            return
+        }
+
+        val newMetric = RemoteMetric(
+            type = "Counter",
```
didiergarcia, line 252: Const?

core/src/main/java/com/segment/analytics/kotlin/core/Telemetry.kt
MichaelGHSeg: There is in the android context plugin.

core/src/main/java/com/segment/analytics/kotlin/core/Telemetry.kt
MichaelGHSeg: Hrm.  Telemetry will be around before Analytics, and Analytics' CoroutineExceptionHandler actually seems like a good place for a Telemetry.error call... which seems incompatible with reusing that scope.  Does making a new scope + handler  sound good?

core/src/main/java/com/segment/analytics/kotlin/core/Telemetry.kt
MichaelGHSeg: It's following the logic in analytics_next.  
Hrm.. if the sampleRate is 0.0 from settings, >= would let it send data (on a 4e15th chance)
But actually, I should add logic to just not start if sample rate is zero, either way.  👍 
... Also because I just realized I'm using sampleRate as the divisor for value.  Glad I looked!  More checks added.

core/src/main/java/com/segment/analytics/kotlin/core/Telemetry.kt
MichaelGHSeg: Apparently it's .lowercase() now, thanks.

core/src/main/java/com/segment/analytics/kotlin/core/Telemetry.kt
MichaelGHSeg: ...  Yes.  Yes it is.  Let's not talk about the 5 minutes I spent staring at this earlier today reassuring myself it made sense.

core/src/main/java/com/segment/analytics/kotlin/core/Telemetry.kt
MichaelGHSeg: That should _definitely_ not be `!=`.  😬

core/src/main/java/com/segment/analytics/kotlin/core/Telemetry.kt
MichaelGHSeg: My C ancestors are disappointed in me.  Thank you.

core/src/main/java/com/segment/analytics/kotlin/core/Telemetry.kt
MichaelGHSeg: If you're pointing out it could be a default value, I'm (ab)using the default value for `log` to have it be left out during `Json.encodeToString`, which complicates things for the other things that never change.  If you have a less hacky solution, I'm all ears.  Def putting some comments around that though, 👍

core/src/main/java/com/segment/analytics/kotlin/core/Telemetry.kt
MichaelGHSeg: Well, now that this happens in start, maybe?  I'll take a look.

MichaelGHSeg: After some sleep - not sure why I assumed it was meant to reuse an existing scope.  👍

core/src/main/java/com/segment/analytics/kotlin/core/Telemetry.kt
```diff
@@ -0,0 +1,271 @@
...
+        val exceptionHandler = CoroutineExceptionHandler { _, t ->
+            errorHandler?.let {
+                it( Exception(
+                    "Caught Exception in Telemetry Scope: ${t.message}",
+                    t
+                ))
+            }
+        }
+        telemetryScope = CoroutineScope(SupervisorJob() + exceptionHandler)
+        telemetryDispatcher = Executors.newCachedThreadPool().asCoroutineDispatcher()
```
wenxi-zeng, line 97: I think we only need one thread for telemetry. maybe `newSingleThread` is better?

core/src/main/java/com/segment/analytics/kotlin/core/Telemetry.kt
```diff
@@ -0,0 +1,271 @@
...
+        }
+
+    private val queue = mutableListOf<RemoteMetric>()
+    private var queueBytes = 0
+    private var queueSizeExceeded = false
+    private val seenErrors = mutableMapOf<String, Int>()
+    private var started = false
+    private var rateLimitEndTime: Long = 0
+    private var telemetryScope: CoroutineScope? = null
+    private var telemetryDispatcher: ExecutorCoroutineDispatcher? = null
```
wenxi-zeng, line 74: we should initialize values for `errorHandler`, `telemetryScope`, and `telemetryDispatcher` when declared, so they don't have to be null. looks like `errorHandler` is preventing us to do this way, but we can create a wrapper for `errorHandler` that invokes errorHanlder if not null, or it does nothing

core/src/main/java/com/segment/analytics/kotlin/core/Telemetry.kt
```diff
@@ -0,0 +1,271 @@
...
+    fun start() {
+        if (started || sampleRate == 0.0) return
+        started = true
+
+        // Assume sampleRate is now set and everything in the queue hasn't had it applied
+        if (Math.random() > sampleRate) {
+            resetQueue()
+        } else {
+            queue.forEach {
+                it.value = (it.value / sampleRate).roundToInt()
```
wenxi-zeng, line 84: curious why do we need to do this here? can we do it when enqueue or dequeue the item? if it's for caching purpose, it's better to do the following so we don't run into concurrent issue:
```
while (queue is not empty) {
    dequeue
    compute the new value
    enqueue again
}
```

core/src/main/java/com/segment/analytics/kotlin/core/Telemetry.kt
```diff
@@ -0,0 +1,271 @@
...
+            }
+        }
+    }
+
+    fun reset()
+    {
+        telemetryDispatcher?.close()
+        telemetryScope = null
+        telemetryDispatcher = null
+        resetQueue()
```
wenxi-zeng, line 122: no need to reset dispatcher and scope. keep a reference to the launched job, cancel that job, then launch a new one when restart. take a look at the [frequency flush policy](https://github.com/segmentio/analytics-kotlin/blob/main/core/src/main/java/com/segment/analytics/kotlin/core/platform/policies/FrequencyFlushPolicy.kt) as an example

core/src/main/java/com/segment/analytics/kotlin/core/Telemetry.kt
```diff
@@ -0,0 +1,271 @@
...
+        seenErrors.clear()
+        started = false
+        rateLimitEndTime = 0
+    }
+
+    fun increment(metric: String, tags: Map<String, String>) {
+        if (!enable || sampleRate == 0.0) return
+        if (!metric.startsWith("analytics_mobile.")) return
```
wenxi-zeng, line 130: let's make `analytics_mobile` a constant

core/src/main/java/com/segment/analytics/kotlin/core/Telemetry.kt
```diff
@@ -0,0 +1,271 @@
...
+        mapOf(
+            "os" to System.getProperty("os.name") + "-" + osVersion,
+            "interpreter" to System.getProperty("java.vendor") + "-" + System.getProperty("java.version"),
+            "library" to "analytics.kotlin",
+            "library_version" to Constants.LIBRARY_VERSION
+        )
+    }
+
+    private fun addRemoteMetric(metric: String, tags: Map<String, String>, value: Int = 1, log: String? = null) {
+        val found = queue.find {
```
wenxi-zeng, line 242: this might be too slow. we should use a map for quick look up

core/src/main/java/com/segment/analytics/kotlin/core/Telemetry.kt
```diff
@@ -0,0 +1,271 @@
...
+            sampleRate = 0.0
+        }
+    }
+
+    private fun send() {
+        try {
+            // Json.encodeToString by default does not include default values
+            //  We're using this to leave off the 'log' parameter if unset.
+            val payload = Json.encodeToString(mapOf("series" to queue))
+            resetQueue()
```
wenxi-zeng, line 196: might have concurrent issue (the flush thread calls this function while the main thread adds an item to the queue). we should snapshot the queue first. in addition, `mutableListOf` will throw concurrent modification exception, should change the type of the queue to a copy on write array, or change it to a real queue that we can dequeue element at the snapshotted size

wenxi-zeng (commented):
overall in good shape. just three things needs to be addressed before merged:
1. possible concurrent issue on the telemetry queue
2. consolidate 3 error reporting methods to 1 for simplicity
3. better to have a lookup table for telemetry queue

core/src/main/java/com/segment/analytics/kotlin/core/Analytics.kt
```diff
@@ -81,6 +81,9 @@ open class Analytics protected constructor(
 
     init {
         require(configuration.isValid()) { "invalid configuration" }
+        Telemetry.increment(Telemetry.INVOKE,
```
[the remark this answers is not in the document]
MichaelGHSeg: For clarification, what's an example of what we want to be consistent with?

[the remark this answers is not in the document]
wenxi-zeng: I lied. I thought analyticsScope is prepared in a separate function not in `build`. let's just move this to the build function 😄

core/src/main/java/com/segment/analytics/kotlin/core/Telemetry.kt
MichaelGHSeg: Could you outline what you're thinking for this?

core/src/main/java/com/segment/analytics/kotlin/core/Analytics.kt
```diff
@@ -121,6 +116,10 @@ open class Analytics protected constructor(
         add(ContextPlugin())
         add(UserInfoPlugin())
 
+        Telemetry.increment(Telemetry.INVOKE_METRIC,
```
wenxi-zeng, line 119: just realize this. we could change the Telemetry functions to be like the following:
```
    fun increment(metric: String, buildTags: (MutableMap<String, String>) -> Unit) {
        val map = mutableMapOf<String, String>()
        buildTags(map)
    }
```
and then the usage becomes much cleaner
```
        increment("metric") {
            it["error"] = "error"
            it["config"] = "config"
        }
```
same applies to the other methods

core/src/test/kotlin/com/segment/analytics/kotlin/core/TelemetryTest.kt
```diff
@@ -89,23 +89,23 @@ class TelemetryTest {
...
+        Telemetry.increment("wrong_metric") { it["test"] = "test" }
         assertEquals(0, TelemetryQueueSize())
         assertEquals(0,errors.size)
     }
 
     @Test
     fun `Test increment with no tags`() {
         Telemetry.start()
-        Telemetry.increment(Telemetry.INVOKE_METRIC, emptyMap())
+        Telemetry.increment(Telemetry.INVOKE_METRIC) { it.clear() }
```
wenxi-zeng, line 108: no need to clear. can just leave it with empty brackets. or make the `tagBuilder` nullable, so you don't have to pass anything

wenxi-zeng (approved):

core/src/test/kotlin/com/segment/analytics/kotlin/core/TelemetryTest.kt
MichaelGHSeg: It was in the test so I just figured I'd make it explicit about what it was expected to do, rather than minimize code.

wenxi-zeng (approved):

PR #217, opened 2024-04-17
Set defaultSettings config option to a var

niallzato:
Despite being a config option when initializing an analytics instance, defaultSettings cannot be assigned due to it being a val. 
As per Kotlin’s docs: https://developer.android.com/kotlin/learn#:~:text=Kotlin%20uses%20two%20different%20keywords,variable%20whose%20value%20can%20change. 

Kotlin uses two different keywords to declare variables: val and var.

Use val for a variable whose value never changes. You can't reassign a value to a variable that was declared using val.

Use var for a variable whose value can change.

image


didiergarcia (approved):

didiergarcia:
LGTM

PR #250, opened 2024-12-03
Improving threading safety for telemetry

MichaelGHSeg:


core/src/main/java/com/segment/analytics/kotlin/core/Telemetry.kt
```diff
@@ -113,8 +121,8 @@ object Telemetry: Subscriber {
      * Called automatically when Telemetry.enable is set to true and when configuration data is received from Segment.
      */
     fun start() {
-        if (!enable || started || sampleRate == 0.0) return
-        started = true
+        if (!enable || started.get() || sampleRate == 0.0) return
```
wenxi-zeng, line 124: if we need atomic on `started`, likely we need it on `enable`

core/src/main/java/com/segment/analytics/kotlin/core/Telemetry.kt
```diff
@@ -76,7 +77,14 @@ object Telemetry: Subscriber {
     var host: String = Constants.DEFAULT_API_HOST
     // 1.0 is 100%, will get set by Segment setting before start()
     // Values are adjusted by the sampleRate on send
-    var sampleRate: Double = 1.0
+    @Volatile private var _sampleRate: Double = 1.0
+    var sampleRate: Double
+        get() = _sampleRate
+        set(value) {
+            synchronized(this) {
```
wenxi-zeng, line 84: we should try our best to avoid using `synchronized` since our API is coroutine-based. people could use our API inside a coroutine which causes a suspend function to block.

core/src/main/java/com/segment/analytics/kotlin/core/Telemetry.kt
```diff
@@ -227,16 +234,14 @@ object Telemetry: Subscriber {
...
-        // Reset queue data size counter since all current queue items will be removed
-        queueBytes = 0
-        val sendQueue = mutableListOf<RemoteMetric>()
-        while (queueCount-- > 0 && !queue.isEmpty()) {
-            val m = queue.poll()
-            if(m != null) {
-                m.value = (m.value / sampleRate).roundToInt()
-                sendQueue.add(m)
-            }
+        val sendQueue: MutableList<RemoteMetric>
```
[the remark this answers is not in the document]
MichaelGHSeg: That's what the code was doing originally, can probably just make the queueBytes atomic and leave it at that.

core/src/main/java/com/segment/analytics/kotlin/core/Telemetry.kt
```diff
@@ -116,16 +112,27 @@ object Telemetry: Subscriber {
...
+        }
+    }
+
     /**
      * Starts the telemetry if it is enabled and not already started, and the sample rate is greater than 0.
      * Called automatically when Telemetry.enable is set to true and when configuration data is received from Segment.
      */
     fun start() {
-        if (!enable || started.get() || sampleRate == 0.0) return
+        if (!enable || started.get() || sampleRate.get() == 0.0) return
```
wenxi-zeng, line 131: should enable be atomic too?

core/src/main/java/com/segment/analytics/kotlin/core/Telemetry.kt
```diff
@@ -334,7 +351,7 @@ object Telemetry: Subscriber {
     private suspend fun systemUpdate(system: com.segment.analytics.kotlin.core.System) {
         system.settings?.let { settings ->
             settings.metrics["sampleRate"]?.jsonPrimitive?.double?.let {
-                sampleRate = it
+                sampleRate.set(it)
```
wenxi-zeng, line 354: if use sovran, sampleRate does not have to be atomic, since sovran serialize the task in the same thread. same thing applies to `enable` and `started`

PR #252, opened 2025-01-23
Customizable storage

wenxi-zeng:


core/src/main/java/com/segment/analytics/kotlin/core/Storage.kt
```diff
@@ -28,6 +29,8 @@ interface Storage {
          * is not present in payloads themselves, but is added later, such as `sentAt`, `integrations` and other json tokens.
          */
         const val MAX_BATCH_SIZE = 475000 // 475KB.
+
+        const val MAX_FILE_SIZE = 475_000
```
MichaelGHSeg, line 33: Do we think these will need to be different?  Maybe good just for variable name clarity.

android/src/main/java/com/segment/analytics/kotlin/android/utilities/AndroidKVS.kt
```diff
@@ -6,10 +6,23 @@ import com.segment.analytics.kotlin.core.utilities.KVS
 /**
  * A key-value store wrapper for sharedPreferences on Android
  */
-class AndroidKVS(val sharedPreferences: SharedPreferences) : KVS {
-    override fun getInt(key: String, defaultVal: Int): Int =
+class AndroidKVS(val sharedPreferences: SharedPreferences): KVS  {
```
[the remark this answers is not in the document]
wenxi-zeng: I think that's fine. people don't have access to the instance of KVS.

core/src/main/java/com/segment/analytics/kotlin/core/Storage.kt
wenxi-zeng: no. it's not gonna be different. it's something set on backend.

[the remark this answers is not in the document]
wenxi-zeng: I think it's better to add a comment to specify the unit. like the one with MAX_BATCH_SIZE. it's cleaner this way and avoiding breaking changes, and is a consistent name across all platforms.

core/src/main/java/com/segment/analytics/kotlin/core/Storage.kt
```diff
@@ -98,11 +101,16 @@ fun parseFilePaths(filePathStr: String?): List<String> {
  *  provider via this interface
  */
 interface StorageProvider {
+    @Deprecated("Deprecated in favor of create which takes vararg params",
+        ReplaceWith("createStorage(analytics, store, writeKey, ioDispatcher, application)")
```
MichaelGHSeg, line 105: We say to replace it with createStorage with all the parameters, but are there any implementations of createStorage that use anything other than the Analytics param?

MichaelGHSeg: That is, should it just be `ReplaceWith("createStorage(analytics)")`

PR #253, opened 2025-01-27
Fix session start data issue

didiergarcia:
We identified an issue with data loss that would occur when the host app would startup with no internet connection. 

In these circumstances the SDK would not be able to download a valid Settings object and would not use either a good default or fallback to the previous known good Settings object.

This fix address various issues that lead to dataloss: 
* Sets the default Configuration.settings object to `null` instead of an empty Settings object (parity with analytics-swit)
* Establish and use good defaults for Settings object when one can not be obtained (parity with analytics-swift)
* Sets the default state of DestinationPlugin.enabled to `true` instead of `false` (parity with intention of analytics-swift)

Note: This PR contains a revert of a previous commit that set Configuration.settings object to an empty Settings object, so there are many changes related to the tests.

### DIFF
--- a/core/src/main/java/com/segment/analytics/kotlin/core/platform/Plugin.kt
+++ b/core/src/main/java/com/segment/analytics/kotlin/core/platform/Plugin.kt
@@ -89,7 +89,7 @@ abstract class DestinationPlugin : EventPlugin {
     override val type: Plugin.Type = Plugin.Type.Destination
     private val timeline: Timeline = Timeline()
     override lateinit var analytics: Analytics
-    internal var enabled = false
+    internal var enabled = true
     abstract val key: String
 
     override fun setup(analytics: Analytics) {


MichaelGHSeg (approved):

core/src/main/java/com/segment/analytics/kotlin/core/State.kt
```diff
@@ -25,14 +25,38 @@ data class System(
...
-                    Settings.serializer(),
-                    storage.read(Storage.Constants.Settings) ?: ""
-                )
-            } catch (ignored: Exception) {
-                configuration.defaultSettings
+            val storedSettings = storage.read(Storage.Constants.Settings)
+            val defaultSettings = configuration.defaultSettings ?: Settings(
+                integrations = buildJsonObject {
+                    put(
+                        "Segment.io",
```
wenxi-zeng, line 32: maybe better to make this a constant in SegmentDestination and reference it here?

wenxi-zeng: I think we can add constructors in Settings so it's consistent with [swift's implementation](https://github.com/segmentio/analytics-swift/blob/main/Sources/Segment/Settings.swift#L18) and that forces every settings created to have a segment destination

core/src/main/java/com/segment/analytics/kotlin/core/platform/Plugin.kt
```diff
@@ -89,7 +89,7 @@ abstract class DestinationPlugin : EventPlugin {
     override val type: Plugin.Type = Plugin.Type.Destination
     private val timeline: Timeline = Timeline()
     override lateinit var analytics: Analytics
-    internal var enabled = false
+    internal var enabled = true
```
wenxi-zeng, line 92: 👍 good catch! otherwise we might have issue when the user first opens the app offline and plugins are disabled by default.

wenxi-zeng: just found out swift actually has defaulted to false. would that cause a behavior difference?

wenxi-zeng (approved):

SebPawp:
Wahou! I can't believe this, we have been waiting for more than a year for this fix! Thank you! 🙏

PR #258, opened 2025-03-10
Add deep link tests

didiergarcia:
Refactored out the code used to extract Uri parameters in DeepLinkUtils so that we could test it.

The tests confirm that we can extract parameters from Hierarchical Uris including the sample stockx.com URI.

### DIFF
--- a/android/src/main/java/com/segment/analytics/kotlin/android/utilities/DeepLinkUtils.kt
+++ b/android/src/main/java/com/segment/analytics/kotlin/android/utilities/DeepLinkUtils.kt
@@ -40,7 +40,7 @@ class DeepLinkUtils(val analytics: Analytics) {
                 put("referrer", it)
             }
 
-            if (uri != null) {
+            uri?.let {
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/segment/analytics/kotlin/android/utilities/DeepLinkUtils.kt
```diff
@@ -13,13 +15,32 @@ class DeepLinkUtils(val analytics: Analytics) {
...
+        referrer: String?,
+        uri: Uri?
+    ): JsonObject {
+        val properties = buildJsonObject {
             referrer?.let {
                 put("referrer", it)
             }
 
-            intent.data?.let { uri ->
+            if (uri != null) {
```
wenxi-zeng, line 43: I think this is the same as the code before?

wenxi-zeng (dismissed):

wenxi-zeng (approved):

PR #281, opened 2025-12-01
Http 2 client

wenxi-zeng:
* replace `HttpURLConnection` with `OkHttp`

### DIFF
--- a/core/src/test/kotlin/com/segment/analytics/kotlin/core/WaitingTests.kt
+++ b/core/src/test/kotlin/com/segment/analytics/kotlin/core/WaitingTests.kt
@@ -96,7 +96,7 @@ class WaitingTests {
     @Test
     fun `test timeout force resume`() = testScope.runTest {
         assertTrue(analytics.running())
-        val waitingPlugin = ManualResumeWaitingPlugin()
+        val waitingPlugin = ExampleWaitingPlugin()
         analytics.add(waitingPlugin)
         analytics.track("foo")
 
@@ -123,7 +123,6 @@ class WaitingTests {
         assertFalse(plugin1.tracked)
         assertFalse(plugin2.tracked)
 
-        plugin1.resume()
         advanceTimeBy(6000)
 
         assertFalse(analytics.running())
@@ -131,8 +130,6 @@ class WaitingTests {
         assertFalse(plugin2.tracked)
 
         plugin2.resume()
-        advanceUntilIdle()
-        advanceTimeBy(6000)
 
         assertTrue(analytics.running())
         assertTrue(plugin1.tracked)


core/src/main/java/com/segment/analytics/kotlin/core/utilities/OkHttpURLConnection.kt
```diff
@@ -0,0 +1,393 @@
...
+
+    // OkHttp-only state management
+    private var _requestMethod = "GET"
+    private var _doInput = true
+    private var _doOutput = false
+    private var _allowUserInteraction = false
+    private var _useCaches = true
+    private var _ifModifiedSince = 0L
+    private var _connectTimeout = 15000
+    private var _readTimeout = 20000
```
didiergarcia: Huh, same values again.

core/src/main/java/com/segment/analytics/kotlin/core/utilities/OkHttpURLConnection.kt
```diff
@@ -0,0 +1,393 @@
...
+            }
+            "PUT" -> {
+                val body = requestBodyBuffer?.readByteArray()?.toRequestBody(
+                    getRequestProperty("Content-Type")?.toMediaType() ?: "text/plain".toMediaType()
+                ) ?: "".toRequestBody("text/plain".toMediaType())
+                builder.put(body)
+            }
+            "DELETE" -> builder.delete()
+            "HEAD" -> builder.head()
+            else -> throw ProtocolException("Unknown method: $_requestMethod")
```
didiergarcia: With Enum does this go away?

core/src/main/java/com/segment/analytics/kotlin/core/utilities/OkHttpURLConnection.kt
```diff
@@ -0,0 +1,393 @@
...
+
+    // Override ALL URLConnection methods to prevent bypassing OkHttp
+    override fun getURL(): URL = url
+
+    override fun getHeaderFieldKey(n: Int): String? {
+        if (!connected) {
+            try { connect() } catch (e: IOException) { return null }
+        }
+        val headers = response?.headers ?: return null
+        return if (n in 0 until headers.size) headers.name(n) else null
```
didiergarcia: This is confusing to me. If headers = null, will the call in the if to headers.size work and not throw a NPE?

core/src/main/java/com/segment/analytics/kotlin/core/utilities/OkHttpURLConnection.kt
```diff
@@ -0,0 +1,393 @@
...
+    override fun setFixedLengthStreamingMode(contentLength: Long) {
+        // OkHttp handles content length automatically, store for compatibility
+    }
+
+    override fun setRequestMethod(method: String) {
+        if (connected) {
+            throw IllegalStateException("Already connected")
+        }
+        _requestMethod = method
+    }
```
didiergarcia: SO internally we can do this (goes from GET to POST), but if someone is using the API tries it should fail; cool.

core/src/main/java/com/segment/analytics/kotlin/core/utilities/OkHttpURLConnection.kt
MichaelGHSeg: certainly not the most verbose way of doing it 😅

core/src/main/java/com/segment/analytics/kotlin/core/HTTPClient.kt
```diff
@@ -170,9 +183,9 @@ open class RequestFactory {
             Analytics.reportInternalError(error)
             throw error
         }
-        val connection = requestedURL.openConnection() as HttpURLConnection
+        val connection = requestedURL.openOkHttpConnection() as HttpURLConnection
         connection.connectTimeout = 15_000 // 15s
-        connection.readTimeout = 20_1000 // 20s
+        connection.readTimeout = 20_000 // 20s
```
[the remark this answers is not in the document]
wenxi-zeng: they could be different values. I wrote it out here so people who use requestFactory know that they could overwrite the values here

core/src/main/java/com/segment/analytics/kotlin/core/utilities/OkHttpURLConnection.kt
wenxi-zeng: yeah. if it's enum, the else branch is not necessary. I could change it to enum, but the `setRequestMethod` takes a string. we'd have to do a conversion internally. besides that users are not going to use the enum, because the connection auto converts GET to POST. they are purely internal.

core/src/main/java/com/segment/analytics/kotlin/core/utilities/OkHttpURLConnection.kt
wenxi-zeng: no. line 165 will already return null if `response?.headers` is null

core/src/main/java/com/segment/analytics/kotlin/core/utilities/OkHttpURLConnection.kt
```diff
@@ -0,0 +1,393 @@
...
+
+    override fun getLastModified(): Long {
+        return getHeaderField("Last-Modified")?.let { dateStr ->
+            try {
+                java.text.SimpleDateFormat("EEE, dd MMM yyyy HH:mm:ss zzz", java.util.Locale.US)
+                    .parse(dateStr)?.time ?: 0L
+            } catch (e: Exception) { 0L }
+        } ?: 0L
+    }
```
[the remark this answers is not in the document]
wenxi-zeng: good call

core/src/main/java/com/segment/analytics/kotlin/core/utilities/OkHttpURLConnection.kt
```diff
@@ -0,0 +1,393 @@
...
+
+    override fun getReadTimeout(): Int = _readTimeout
+
+    override fun toString(): String {
+        return "OkHttpURLConnection:$url"
+    }
+
+    // Override ALL HttpURLConnection specific methods
+    override fun getRequestMethod(): String = _requestMethod
+
```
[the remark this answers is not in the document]
wenxi-zeng: no. string is immutable. reassignment only changes the local variable, not the original one

core/src/main/java/com/segment/analytics/kotlin/core/utilities/OkHttpURLConnection.kt
```diff
@@ -0,0 +1,393 @@
...
+    }
+
+    override fun getContent(classes: Array<out Class<*>?>?): Any? {
+        if (!connected) connect()
+
+        val content = getContent()
+        if (classes == null || classes.isEmpty()) {
+            return content
+        }
+
```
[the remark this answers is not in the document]
didiergarcia: So the null check has a side-effect of casting? Well that's annoying... Understood.

PR #284, opened 2025-12-16
Fix file descriptor leak in EventPipeline

apuig:
Changed from `.let` to `.use` to ensure `InputStream` is always closed, preventing FD exhaustion when exceptions occur during upload.

This results in serious problems when environments are unable to connect to the Segment API over an extended duration


apuig:
Hi @wenxi-zeng 
We’ve encountered this issue and would appreciate your help. 
Thank you in advance!

PR #287, opened 2026-01-15
Fix Windows file deletion by closing stream before removeFile

apuig:
Follow-up to #284

Move `storage.removeFile()` call outside the `use{}` block to ensure `InputStream` is fully closed before attempting file deletion. On Windows, files cannot be deleted while file handles remain open, causing silent deletion failures and infinite upload retries of the same batch files.


wenxi-zeng:
@apuig forgot to get back to you. this fix has been released in `1.24.1`

PR #297, opened 2026-02-27
Fix blocking http calls which cannot be cancelled

olamy:
The analytics-kotlin library runs blocking OkHttp calls (client.newCall(request).execute()) inside coroutines via withContext and launch, but without runInterruptible. This means Job.cancel() cannot interrupt in-flight HTTP calls the coroutine only cancels at the next suspension point, not during the blocking I/O. Wrapping in runInterruptible causes coroutine cancellation to trigger Thread.interrupt(), which OkHttp handles by aborting the connection immediately.

Signed-off-by: Olivier Lamy <olamy@apache.org>


wenxi-zeng (approved):

apuig:
Hi @wenxi-zeng ! Since this PR has been approved, is there an ETA for cutting a release that will include it? Thanks!

PR #302, opened 2026-03-20
Fix SDK bugs found by e2e retry tests

MichaelGHSeg:
## Summary
Fixes found while running e2e retry test suite against the smart retry pipeline:

**EventPipeline fixes:**
- Smart retry override: `RetryStateMachine.shouldDeleteBatch()` replaces legacy `handleUploadException` for cleanup decisions — aligns with `statusCodeOverrides` (410/460 now retry, 501/505 now drop)
- Case-insensitive Retry-After header lookup (OkHttp/HTTP2 lowercases response headers)
- X-Retry-Count omitted on first attempt, sent only on retries (1, 2, 3...)
- DropBatch decision now fires `reportInternalError` so errorHandler can detect dropped batches

**RetryStateMachine:**
- New `shouldDeleteBatch()` public method for EventPipeline cleanup decisions

**e2e-cli:**
- Wire `errorHandler` to detect delivery failures (HTTP errors, dropped batches)
- Wire `maxRetries` from test config to SDK `BackoffConfig`
- Clear delivery errors before each flush cycle
- Report failure when retries exhausted or batch dropped

## Test plan
- [x] 48/48 e2e tests pass (basic + retry suites)
- [x] Existing unit tests unaffected

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -14,10 +14,17 @@ on:
     branches: [main, master]
   pull_request:
     branches: [main, master]
-  workflow_dispatch: # Allow manual trigger
+  workflow_dispatch:
+    inputs:
+      e2e_tests_ref:
+        description: 'Branch or ref of sdk-e2e-tests to use'
+        required: false
+        default: 'main'
 
 jobs:
   e2e-tests:
+    # Skip on fork PRs where repo secrets aren't available
+    if: ${{ github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository }}
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/segment/analytics/kotlin/core/platform/EventPipeline.kt
```diff
@@ -233,20 +237,31 @@ open class EventPipeline(
...
                         analytics.reportInternalError(e)
-                        
+
                         // Extract status code and retry-after from exception
                         if (e is HTTPException) {
                             statusCode = e.responseCode
-                            retryAfterSeconds = e.responseHeaders["Retry-After"]?.firstOrNull()?.toIntOrNull()
+                            // Header names may be lowercased by HTTP/2 or OkHttp
+                            retryAfterSeconds = (e.responseHeaders["Retry-After"]
+                                ?: e.responseHeaders["retry-after"])
```
MichaelGHSeg: Should we maybe just only be checking lowercase here?

didiergarcia: yeah, .toLowerCase() and one check?

PR #313, opened 2026-06-17
feat: honor Retry-After header on any retryable HTTP response

didiergarcia:
## Summary
- Any retryable HTTP response carrying a `Retry-After` header now triggers a pipeline-level pause, identical to 429 behavior
- Codes without `Retry-After` continue to use existing exponential backoff unchanged
- Emits a warning log when `Retry-After` triggers a new pipeline pause (any code including 429)
- Fixes `shouldDeleteBatch()` bug: retryable batches were incorrectly deleted when `rateLimitConfig.enabled=true, backoffConfig.enabled=false`
- Existing 429 behavior is unchanged

## Motivation
Per RFC 9110, `Retry-After` is a general-purpose header. Previously the SDK only respected it for 429; a 529 (or 503) with `Retry-After` would be silently ignored and the SDK would compute its own (potentially shorter) backoff, increasing load on an already-overloaded server.

## Test plan
- [ ] `RetryAfterGenericTest` — all 13 new tests pass
- [ ] Full Gradle test suite passes with no regressions in retry tests
- [ ] Manual: connect to test endpoint returning 529 + `Retry-After: 5`, observe pipeline pause in logs

### DIFF
--- a/core/src/main/java/com/segment/analytics/kotlin/core/retry/RetryStateMachine.kt
+++ b/core/src/main/java/com/segment/analytics/kotlin/core/retry/RetryStateMachine.kt
@@ -45,11 +45,19 @@ class RetryStateMachine(
                 }
             }
 
-            behavior == RetryBehavior.RETRY && config.backoffConfig.enabled -> {
-                handleRetryableError(state, response, currentTime)
+            behavior == RetryBehavior.RETRY -> {
+                when {
+                    response.retryAfterSeconds != null && config.rateLimitConfig.enabled -> {
+                        handleRateLimitResponse(state, response, currentTime)
+                    }
+                    config.backoffConfig.enabled -> {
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/segment/analytics/kotlin/core/retry/RetryStateMachine.kt
```diff
@@ -189,8 +197,11 @@ class RetryStateMachine(
         // 429 with rate limit handling disabled: delete
         if (statusCode == 429) return !config.rateLimitConfig.enabled
```
MichaelGHSeg, line 198: Maybe redundant with the new code below?  429 probably just not a unique case anywhere now.

MichaelGHSeg (approved):

